PR #38, opened 2025-04-10
implement state management within the flow

Daan0709:
- Flow now makes use of a state manager for easier management of the nodes in the future


philipsens (approved):
Currently, there is one store for keeping state of the flow.
In the future it should be considered if this state will get reset on flow change or if multiple states are going to be kept. 

We still need a FileStore, that could fill the FlowStore on the fly, or it could have a FlowState for each open file.

PR #42, opened 2025-04-18
Feat/drag n drop

### LINKED ISSUE #40: Drag and drop nodes from sidebar
Nodes should be draggable from the sidebar onto the canvas

Daan0709:
- Can now drag nodes from sidebar onto the canvas
- Will change the sidebar to a form in which the user can fill out all the required attributes to be displayed on the node

Closes #40

### DIFF
--- a/src/main/react/app/routes/builder/context/node-context.tsx
+++ b/src/main/react/app/routes/builder/context/node-context.tsx
@@ -38,7 +38,7 @@ export default function NodeContext({
   function resolveFilledAttributes() {
     const filledAttributes = Object.entries(inputReferences.current)
       .map(([indexString, input]) => {
-        const index = Number(indexString)
+        const index = +indexString
         const value = input?.value?.trim()
         if (value) {
           return {


src/main/react/app/routes/builder/context/node-context.tsx
```diff
@@ -0,0 +1,110 @@
...
+export default function NodeContext({
+  nodeId,
+  setShowNodeContext,
+}: Readonly<{
+  setShowNodeContext: (b: boolean) => void
+  nodeId: number
+}>) {
+  const attributes = useNodeContextStore((state) => state.attributes)
+  const inputReferences = useRef<Record<number, HTMLInputElement | null>>({})
+  const { setAttributes, setNodeName } = useFlowStore((state) => state)
```
philipsens, line 14: I don't think this arrow function in needed when selecting the whole state.
Maybe selecting stuff isn't a bad idea, to minimize component re-renders. Otherwise, we could probably do a refactor afterward. (Thinking of it, re-renders might not really be a problem in this component, but still a good practice)

src/main/react/app/routes/builder/context/node-context.tsx
```diff
@@ -0,0 +1,110 @@
...
+      setCanSave(true)
+      return
+    }
+    validateForm()
+  }, [attributes])
+
+  const handleSave = () => {
+    const filledAttributes = Object.entries(inputReferences.current)
+      .map(([indexString, input]) => {
+        const index = Number(indexString)
```
philipsens, line 40: We're more a fan of this casting method (keeping it a primitive):
```suggestion
        const index = +indexString
```

src/main/react/app/routes/builder/context/node-context.tsx
```diff
@@ -0,0 +1,110 @@
...
+  useEffect(() => {
+    if (!attributes) {
+      setCanSave(true)
+      return
+    }
+    validateForm()
+  }, [attributes])
+
+  const handleSave = () => {
+    const filledAttributes = Object.entries(inputReferences.current)
```
philipsens, line 38: It is probably better for readability for `handleSave` to separate this into a method.

src/main/react/app/routes/builder/node-translator-module.ts
```diff
@@ -0,0 +1,22 @@
+const typeSuffixes = ['pipe', 'listener', 'receiver', 'sender', 'validator', 'wrapper', 'job', 'exit']
+
+const typeColorMap: Record<string, string> = {
```
philipsens, line 3: It might be better to declare these colors as variables in CSS.

philipsens (changes requested):

philipsens (approved):

PR #49, opened 2025-04-25
Implementation of a sticky note on the canvas

### LINKED ISSUE #44: Add 'note' node
Users should have the options to place notes on the canvas. These can be used as personal reminders or just descriptive statements.

These will not have any handles attached to them and will just float seperately on the canvas.

Daan0709:
- By clicking the right mouse button, the user can spawn in a sticky note on top of the canvas

Closes #44


src/main/react/app/routes/builder/context/node-context.tsx
```diff
@@ -26,6 +26,20 @@ export default function NodeContext({
...
   }
 
+  // Fills out input fields with already existing attributes when editing a node
+  useEffect(() => {
+    const currentAttributes = getAttributes(nodeId.toString())
+    if (currentAttributes && attributes) {
+      for (const [index, attribute] of attributes.entries()) {
+        const value = currentAttributes[attribute.name]
+        if (value && inputReferences.current[index]) {
+          inputReferences.current[index]!.value = value
```
philipsens, line 36: Is the non-null assertion operator really needed here? It should only be used in situations where it is not possible to fix otherwise, like in local variables.

```suggestion
        const currentInputReferance = inputReferences.current[index]
        if (value && currentInputReferance) {
          currentInputReferance.value = value
```

src/main/react/app/stores/flow-store.ts
```diff
@@ -60,6 +72,22 @@ const useFlowStore = create<FlowState>((set, get) => ({
...
+  },
+  addNode: (newNode: FlowNode) => {
+    set({
+      nodes: [...get().nodes, newNode],
+    })
+  },
+  deleteNode: (nodeId: string) => {
+    set({
+      nodes: get().nodes.filter((node) => node.id !== nodeId),
+      edges: get().edges.filter((edge) => edge.source !== nodeId && edge.target !== nodeId),
```
philipsens, line 88: These things need to be an array for the ReactFlow library?

src/main/react/app/routes/builder/context/node-context.tsx
Daan0709: Agreed! Oversight on my end

src/main/react/app/stores/flow-store.ts
philipsens: Yes I saw. I was just wondering, because otherwise stuff could be deleted by doing something like: 
```ts
delete node[id]
``` 
Without having to iterate over the list. But ok if it is a library limitation.

src/main/react/app/routes/builder/context/node-context.tsx
philipsens: No problem.

philipsens (approved):

philipsens:
Conflicts :(

PR #54, opened 2025-05-07
Builder context revamped and search function implementation

### LINKED ISSUE #53: Implement search feature on builder-context
In order for users to properly find their elements, a search feature should be implemented within the builder context.

# Tasks
- [ ] Create seperate zones for pipes, senders, listeners, validators and other elements
- [ ] Make these zones collapsable
- [ ] Implement search feature to only show matching hits from those zones, and collapse them if there are no hits

Daan0709:
- Can now search the palette for specific elements
- Palette is now split up in multiple blocks of same-type elements (pipes, senders, listeners etc.)
- Implemented dagre layout for automatic layering of the nodes instead of manual assignment of x and y coordinates

Closes #53


src/main/react/app/routes/builder/context/builder-context.tsx
```diff
@@ -36,24 +90,61 @@ export default function BuilderContext({ onClose }: Readonly<{ onClose: () => vo
...
-                draggable
-                onDragStart={onDragStart(value)}
-              >
-                {value.name}
-              </li>
-            ))}
+            Object.entries(elementsToRender)
+              .sort(([typeA], [typeB]) => typeA.localeCompare(typeB))
+              .map(([type, items]) => {
+                const isExpanded = searchTerm ? true : (expandedGroups[type] ?? true)
```
philipsens, line 109: Does this logic really need to be like this?
Wouldn't this produce the same result?
```suggestion
                const isExpanded = expandedGroups[type] ?? searchTerm
```

Is there some reason there should be changed on searchTerm first?

src/main/react/app/routes/builder/context/builder-context.tsx
```diff
@@ -36,24 +90,61 @@ export default function BuilderContext({ onClose }: Readonly<{ onClose: () => vo
...
+                    className="mb-4 border-t-2 p-2 shadow-md"
+                    style={{
+                      borderColor: `var(--type-${type}`,
+                    }}
+                  >
+                    <button
+                      onClick={() => toggleGroup(type)}
+                      className="w-full cursor-pointer text-left text-sm font-semibold text-gray-600 capitalize hover:text-gray-900"
+                    >
+                      {isExpanded ? '▾' : '▸'} {type === 'other' ? type : `${type}s`}
```
philipsens, line 122: Don't you want to use the icons?

philipsens (changes requested):
I think it would be nice to separate the results into its own component. In the future, it will get even more logic.

src/main/react/app/routes/builder/context/builder-context.tsx
Daan0709: Changed! Had not thought of them

src/main/react/app/routes/builder/context/builder-context.tsx
Daan0709: The purpose of this code is to auto expand all groups when a user has entered something into the searchfield. In your suggestion the group would not be expanded if the toggle for it is false, even if a user would then enter something in the field. 
I felt like that could make users miss certain elements if they did not think to toggle the specific group where it would be located in

philipsens: Ah, so maybe this then:
```ts
searchTerm || expandedGroups[type] ?? true
```

philipsens (approved):

PR #59, opened 2025-05-15
Add a Help page that shows MDX documentation

philipsens:
I've added a simple Help page with a sidebar for topics. The topics can be entered in the `help-topic-tree-items.ts`

The documentation itself composes of multiple MDX files, for easy documenting.

PR #60, opened 2025-05-15
Refactor sidebars into a component

philipsens:
Creates a component that can be used to apply the layout with 2 (or 1) sidebar(s).

Dependent on #59


philipsens:
Because we squash, there is no linear git history and this dependent branch won't work... I'll have to figure this out :(

PR #63, opened 2025-05-19
Hotfix pages not loading when layout store has not been initialized

philipsens:
### Description

- Fixed an issue where pages would not load if the layout store had not been initialized.


philipsens:
We discussed this can be merged in person.

PR #135, opened 2025-11-24
Unit tests

### LINKED ISSUE #122: Create Unit Tests

Daan0709:
- Added tests for classes, services and controllers
- Improved BE clarity with custom exceptions which get caught by the controller

Closes #122


src/main/java/org/frankframework/flow/configuration/ConfigurationDTO.java
Matthbo: For consistency, maybe DTOs should be record classes where possible

Matthbo (approved):

PR #153, opened 2025-12-10
Improve search function

Daan0709:
- Made a single file-structure class for both the editor and the studio to use
- You can now tab through search results and open selected adapters with 'enter'


stijnpotters1 (approved):
Kijk nog even naar de SonarQube warnings

PR #159, opened 2025-12-12
Recursive child elements

### LINKED ISSUE #139: Childelements should also be able to hold child elements
Currently only top-level nodes can hold child elements. Child elements should be able to hold other child elements aswell.

### LINKED ISSUE #145: Translate nested children from Flow to XML

### LINKED ISSUE #156: Validate dragging children into elements
Elements should not be blindly accepting any child being dragged onto them. It should first be validated with the frankdoc to only allow accepted children into the element.

Here are the elements a SenderPipe should accept for example:

Image

Daan0709:
- Child elements are now recursive, meaning they in theory can be stacked infinitely
- Nodes now validate the dropped element against their children property, making it so they no longer blindly accept anything and now only accept valid children
- Added clear indicators when starting a drag from the node-context of where said dragged element can be dropped
- Changed how selected nodes look (instead of thick black border -> thin blue border)

Closes #156, Closes #139, Closes #145

### DIFF
--- a/src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
+++ b/src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
@@ -23,7 +23,7 @@ import { ChildNode } from './child-node'
 import { findChildRecursive } from '~/stores/child-utilities'
 import { canAcceptChildStatic } from './node-utilities'
 
-export type frankNode = Node<{
+export type FrankNodeType = Node<{
   subtype: string
   type: string
   name: string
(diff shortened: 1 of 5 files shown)


src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
```diff
@@ -19,16 +19,11 @@ import variables from '../../../../../environment/environment'
...
-export interface ChildNode {
-  id: string
-  subtype: string
-  type: string
-  name?: string
-  attributes?: Record<string, string>
-}
-
-export type FrankNode = Node<{
+export type frankNode = Node<{
```
Matthbo: Hmm maybe calling it `FrankNodeType` would be better than putting the name into camel-case

src/main/frontend/app/stores/flow-store.ts
```diff
@@ -249,38 +252,48 @@ const useFlowStore = create<FlowState>((set, get) => ({
...
+  addChildToChild: (nodeId: string, targetChildId: string, newChild: ChildNode) => {
+    set({
+      nodes: get().nodes.map((node) => {
+        if (node.id !== nodeId) return node
+
+        return {
+          ...node,
+          data: {
+            ...node.data,
+            children: addChildRecursive(node.data.children || [], targetChildId, newChild),
```
Matthbo: Don't forget to use `??` instead of `||` in order to prevent falsely values triggering unwanted behaviour, I know that won't be the case here but its better to be consistent with it

Matthbo (approved):

PR #162, opened 2025-12-17
Feat/seperate editor file structure

Daan0709:
- The editor now shows the true root structure of the project, while the studio takes src/main/configurations as its root folder
- Seperated the tabs component into studiotabs and editortabs (studiotabs would show adapters, editortabs entire configurations)

### DIFF
--- a/src/main/frontend/app/components/file-structure/editor-data-provider.ts
+++ b/src/main/frontend/app/components/file-structure/editor-data-provider.ts
@@ -47,10 +47,9 @@ export default class EditorFilesDataProvider implements TreeDataProvider {
     this.data[item.index].data.name = name
   }
 
-  // eslint-disable-next-line sonarjs/cognitive-complexity
   private buildTree(rootName: string, paths: string[]) {
     const newData: Record<TreeItemIndex, TreeItem<FileNode>> = {
-      ['root']: {
+      root: {
         index: 'root',
         data: {
           name: rootName,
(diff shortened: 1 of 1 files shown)


src/main/frontend/app/components/file-structure/editor-file-structure.tsx
```diff
@@ -0,0 +1,239 @@
...
+        setActiveMatchIndex(0)
+        setHighlightedItemId(matches[0])
+      } else {
+        setActiveMatchIndex(-1)
+        setHighlightedItemId(null)
+      }
+    }
+
+    findMatchingItems()
+  }, [searchTerm, filepaths])
```
Matthbo:
```suggestion
	// findMatchingItems
  useEffect(async () => {
    if (!searchTerm || !dataProviderReference.current) {
      setMatchingItemIds([])
      setActiveMatchIndex(-1)
      setHighlightedItemId(null)
      return
    }

    const allItems = await dataProviderReference.current.getAllItems()
    const lower = searchTerm.toLowerCase()

    const matches = allItems
      .filter((item) => getItemTitle(item).toLowerCase().includes(lower))
      .map((item) => item.index as string)

    setMatchingItemIds(matches)

    if (matches.length > 0) {
      setActiveMatchIndex(0)
      setHighlightedItemId(matches[0])
    } else {
      setActiveMatchIndex(-1)
      setHighlightedItemId(null)
    }
  }, [searchTerm, filepaths])
```

PR #174, opened 2026-01-09
Read & write from filesystem

Daan0709:
Kind of a large one:
- Users can now add a folder from a specified path on the local drive, which will then get treated as the "project" or root folder. The folder in which the user has their projects stored can be defined in application-local.properties, in the app.project.root property (e.g.: ```app.project.root=C:/Users/User/FlowProjects```)
- The filetree in the editor now directly mimics the folder structure of the one located on the users drive
- The filetree in the studio then treats src/main/configurations as the root, and treats .xml files as folders which then contain adapters as files
- Reading & writing now happens directly off the users drive, instead of on a cloned in-memory filesystem

### DIFF
--- a/src/main/frontend/app/components/file-structure/files-data-provider.ts
+++ b/src/main/frontend/app/components/file-structure/files-data-provider.ts
@@ -16,87 +16,86 @@ export default class FilesDataProvider implements TreeDataProvider {
 
   /** Update the tree using a backend fileTree */
   public async updateData(fileTree: FileTreeNode) {
-    const newData: Record<TreeItemIndex, TreeItem> = {
-      root: {
-        index: 'root',
-        data: 'Configurations',
-        children: [],
-        isFolder: true,
-      },
-    }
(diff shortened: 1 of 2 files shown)


src/main/frontend/app/components/file-structure/files-data-provider.ts
```diff
@@ -57,51 +25,103 @@ export default class FilesDataProvider implements TreeDataProvider {
       },
     }
 
-    for (const { configPath, adapters } of configs) {
-      // Remove the fixed src/main/configurations prefix
-      const relativePath = configPath.replace(/^src\/main\/configurations\//, '')
+    // Recursive traversal
+    const traverse = async (node: FileTreeNode, parentIndex: TreeItemIndex): Promise<TreeItemIndex | null> => {
```
Matthbo: How is this traverse function different from the one in editor-data-provider.ts?
If it's almost identical, maybe think of a way to make that one more generic so it can be reused here. If that's possible

PR #176, opened 2026-01-13
Optimize filetree request path

### LINKED ISSUE #173: Cache file structures for quicker loading
Currently file structures do not get stored anywhere after loading the first time. Caching it in a store would greatly improve loadtimes when revisiting the pages.

Daan0709:
- Now only requests folder contents when actually loading the folder. 
- Sorts the filetree by directory first, then .xml files treated like folders, and then other files

Closes #173

### DIFF
--- a/src/main/frontend/app/routes/app-layout.tsx
+++ b/src/main/frontend/app/routes/app-layout.tsx
@@ -1,8 +1,41 @@
 import Navbar from '~/components/navbar/navbar'
 import { FrankDocProvider } from '~/providers/frankdoc-provider'
 import AppContent from '~/components/app-content'
+import { useEffect, useState } from 'react'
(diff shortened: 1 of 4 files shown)


src/main/frontend/app/components/file-structure/editor-data-provider.ts
```diff
@@ -20,16 +22,40 @@ export default class EditorFilesDataProvider implements TreeDataProvider {
...
   constructor(projectName: string) {
     this.projectName = projectName
-    this.loadRoot()
   }
 
+  /**
+   * Public method to initialize data loading.
+   * Call this from your React component's useEffect.
+   */
+  public async loadData(): Promise<void> {
```
Matthbo: Why not just make `loadRoot` public?
There is no check of root has already been initialized so you could abuse it in the exact same way as you would if the function was public

PR #183, opened 2026-01-14
Xml translation and bugfixes

### LINKED ISSUE #140: XML-to-flow translation is not accurate
The translation of xml to a flow element is not 100% accurate. Here are some configurations that do not translate properly:
- Within a pipeline: If an element also has a Forward element in it, it wont also connect to the next element in the pipeline

expand as new issues get discovered...

### LINKED ISSUE #180: Dragging elements onto canvas does not remove 'can drop here' marker on other elements
Image

### LINKED ISSUE #181: Duplicate 'success' forward
If an element already has a 'success' forward type from the doc, it ends up having two of them

Image

Daan0709:
- Fixed issue where 'succes' handletypes could be duplicated
- Fixed issue where 'can drop here' text would linger on elements if the dragged element was dropped onto the canvas
- Improved xml to flow translation cases, such as issues with edge generations

All adapters must follow the rules below when generating edges (The links between nodes):

## 1. Explicit Forward Edges
- Every `<Forward>` element **must always** generate an edge.
- The edge connects the current node (source) to the target specified in the `<Forward>` element.
- The type and index of the source handle should reflect the order of the forwards.

## 2. Implicit Pipeline Edges
- For nodes within the `<Pipeline>`, edges are generated **top-to-bottom** to simulate the natural flow.
- **Skip creating an implicit edge** in the following cases:
  1. The current node has one or more `<Forward>` elements that target **any other pipeline node**.
  2. The current node has a `<Forward>` element targeting an **exit node with `state="SUCCESS"`**.
- Implicit edges ensure that pipeline execution flows correctly even if explicit forwards are missing, but never override explicit logic.

## 3. Implicit Success Exit Edge
- If there is an **exit node with `state="SUCCESS"`**:
  - And the **last node in the pipeline** does **not** already forward to any exit,
  - Then automatically generate an implicit edge from that last node to the success exit.
- This guarantees that successful pipeline completions are always connected to the success exit.

Closes #180 , Closes #181, Closes #140


Matthbo (approved):

stijnpotters1 (changes requested):
Nog wat feedback

stijnpotters1 (approved):

PR #203, opened 2026-02-04
Added Jump buttons

stijnpotters1:
image

image

### DIFF
--- a/src/main/frontend/app/routes/studio/studio.tsx
+++ b/src/main/frontend/app/routes/studio/studio.tsx
@@ -57,7 +57,7 @@ export default function Studio() {
           <>
             <div className="border-b-border bg-background flex h-12 items-center justify-between border-b p-4">
               <span>Path: {activeTabPath}</span>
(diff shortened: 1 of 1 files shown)


src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -251,8 +308,16 @@ export default function CodeEditor() {
         </div>
         {activeTabFilePath ? (
           <>
-            <div className="border-b-border bg-background flex h-12 items-center border-b p-4">
-              Path: {activeTabFilePath}
+            <div className="border-b-border bg-background flex h-12 items-center justify-between border-b p-4">
+              <span>Path: {activeTabFilePath}</span>
+              <button
```
philipsens: Why not use the button component?

Daan0709:
- Buttons now make use of the navigationActions logic that was also used by the configuration tiles.
- Updated Button component to have a pointer cursor
- Every source of opening a tab now provides the same tabdata to remove inconsistencies and accidental duplicated tabs with slightly different names

PR #206, opened 2026-02-04
Implemented Persistent Filesystem solution

stijnpotters1:
It works both locally and on cloud
Full integration with file system
Open and saves original folder and files on the frank flow.
you can create a new project in your file system and work in it.
you can clone a github repository to your local file system and work in it.
Your recent projects are being tracked and you also can delete then from your recents.

image
image
image
image
image

### DIFF
--- a/src/main/java/org/frankframework/flow/filesystem/FileSystemStorage.java
+++ b/src/main/java/org/frankframework/flow/filesystem/FileSystemStorage.java
@@ -22,7 +22,7 @@ public interface FileSystemStorage {
     void writeFile(String path, String content) throws IOException;
 
     /**
-     * MMakes new folder in directory
+     * Makes new folder in directory
      */
     Path createProjectDirectory(String path) throws IOException;
 


Daan0709 (changes requested):
Some small changes needed, and you should take a look at the security issues.

src/main/frontend/app/components/directory-picker/directory-picker.tsx
```diff
@@ -0,0 +1,158 @@
...
+            >
+              Cancel
+            </button>
+            <button
+              onClick={() => onSelect(activePath)}
+              disabled={!activePath}
+              className="bg-backdrop hover:bg-background border-border cursor-pointer rounded border px-3 py-1 text-sm disabled:cursor-not-allowed disabled:opacity-50"
+            >
+              Select
+            </button>
```
Daan0709: You could use the Button component for these buttons. Makes it easier to apply style changes in the future to buttons

src/main/java/org/frankframework/flow/filesystem/FileSystemStorage.java
```diff
@@ -0,0 +1,39 @@
...
+     * Returns what directory entails
+     */
+    List<FilesystemEntry> listDirectory(String path) throws IOException;
+
+    String readFile(String path) throws IOException;
+
+    void writeFile(String path, String content) throws IOException;
+
+    /**
+     * MMakes new folder in directory
```
Daan0709:
```suggestion
     * Makes new folder in directory
```
Typo 😎

Daan0709 (approved):

PR #216, opened 2026-02-11
Create build-release.yml

VincentvanMedenbach:
The file is almost entirely the same as the one provided by sergi, only changes are the name, and 
  file: docker/Dockerfile 
  added to the docker build, since the dockerfile is located at docker/Dockerfile

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -10,8 +10,6 @@ on:
     branches: [ "master" ]
     # Publish semver tags as releases.
     tags: [ 'v*.*.*' ]
-  pull_request:
-    branches: [ "master" ]
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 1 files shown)


.github/workflows/build-release.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      # Login against a Docker registry except on PR
+      # https://github.com/docker/login-action
+      - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
stijnpotters1: Weet je of deze dan aan de github action secrets toevoegen, of is dit een default token is in GitHub zelf?

PR #229, opened 2026-02-13
Undo & redo

Daan0709:
- You can now undo changes made to the canvas with CTRL + Z 
- You can also redo these undo changes with CTRL + SHIFT + Z or CTRL + Y
- Changes include but are not limited to: moving nodes, deleting nodes, deleting edges, adding nodes, grouping nodes
- A maximum of 20 changes is tracked, this is just a number that I chose cause it felt right (not too much for memory overload, but also a fair amount that can be undone)


stijnpotters1 (approved):
Goed, verder nog één comment

PR #249, opened 2026-02-19
Refactor session handling

stijnpotters1:
Now cloud connection goes through localstorage instead of session storage so you cannot lose connection with your cloud workspace after the session is closed


Daan0709:
> > > I've noticed X-Workspace-ID header is being set for API requests to the backend. What's the purpose of this, cant a jsession be set up through Spring?
> > 
> > 
> > I changed the header because previously i uses the Session header to connect a client session to a workspace in the container. But since i changed it from session to localstorage i though it was unclear when kept the Session header because it was not tracked by sessions anymore, so i refactored it to the X-Workspace-ID header to specify which workspace a client applies to.
> > So yes i could use jsession but i thought it made things unclear. What do you think? @Matthbo
> 
> But doesn't Spring sessions use a jsession cookie instead of a header? If I look at the FF Console for example, it keeps track of the session purely through the cookie which is better since all cookies are always shared with both server and client with every request.

I feel like it is set up this way because the session is not really being tracked here, so using jsession would be the wrong tool. Instead a custom (X-Workspace-ID) header is created to keep track of what workspace belongs to what client. That way there is no need for diving into session infrastructure. Correct me if I'm wrong though @stijnpotters1

PR #268, opened 2026-02-24
Various fixes

### LINKED ISSUE #259: Filter out empty configurations and .xml files that are not configurations from the configuration manager

### LINKED ISSUE #260: Remove '+' handle when an element already has a maximum amount of handles

Daan0709:
- Add handle button is now removed if all available forward types of an element already have a handle.
- If a custom forward/handle is possible, the add handle button stays regardless if a custom handle is already present.
- .xml files without any adapters in them now no longer show up in the configuration manager page.
- Added a search function to the configuration manager page
- Fixed an issue with the projectlanding action buttons looking weird on lower resolution screens

Closes #259, Closes #260

### DIFF
--- a/src/main/frontend/app/routes/configurations/configuration-manager.tsx
+++ b/src/main/frontend/app/routes/configurations/configuration-manager.tsx
@@ -133,6 +133,8 @@ export default function ConfigurationManager() {
 
     let cancelled = false
 
+    // Runs asynchronously in the background to load adapter names
+    // without blocking the UI. State is updated once all files are processed.
     const loadAdapters = async () => {
       const results = await Promise.all(
         configFiles.map(async (file) => {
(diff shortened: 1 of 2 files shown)


src/main/frontend/app/routes/configurations/configuration-manager.tsx
```diff
@@ -106,7 +111,66 @@ export default function ConfigurationManager() {
...
+
+  useEffect(() => {
+    if (!currentProject?.name || configFiles.length === 0) {
+      setFilesWithAdapters([])
+      return
+    }
+
+    let cancelled = false
+
+    const loadAdapters = async () => {
```
Matthbo: Please put a comment on this saying it'll run in the background

PR #279, opened 2026-02-26
Fix for studio bugs

Daan0709:
- Adapters saved from studio are now no longer overwritten by a static ```<Adapter>```, instead keeping their attributes so the editor and studio are in sync
-  Editor tabs can no longer be duplicated by originating one from the file-structure, and one from the jump to editor buttons

### DIFF
--- a/src/main/frontend/app/components/file-structure/editor-data-provider.ts
+++ b/src/main/frontend/app/components/file-structure/editor-data-provider.ts
@@ -24,7 +24,18 @@ export default class EditorFilesDataProvider implements TreeDataProvider {
 
   constructor(projectName: string) {
     this.projectName = projectName
-    void this.fetchAndBuildTree()
+  }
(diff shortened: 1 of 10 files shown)


src/main/java/org/frankframework/flow/project/ProjectService.java
```diff
@@ -291,7 +319,6 @@ public boolean updateConfigurationXml(String projectName, String filepath, Strin
                 .orElseThrow(() -> new ConfigurationNotFoundException(
                         String.format("Configuration with filepath: %s not found", filepath)));
 
-        fileSystemStorage.writeFile(filepath, xmlContent);
```
stijnpotters1: Wat is de reden dat deze line is weg gehaald?
Nu zou dit niet meer de configuratie op het originele filepath updaten

PR #303, opened 2026-03-03
Update palette

Daan0709:
- Fixed visual issue with the `>` or `V` icons disappearing
- Hovering over an element now shows information such as labels and the description
afbeelding

- These will lock in place after the user has hovered over the element for 1 second, allowing the user to scroll within said hovercard. Clicking anywhere outside of it or hovering over a new element will make the card disappear again
- Added a link to the frankdoc for each element

### DIFF
--- a/src/main/frontend/app/actions/navigationActions.ts
+++ b/src/main/frontend/app/actions/navigationActions.ts
@@ -2,18 +2,21 @@ import useTabStore from '~/stores/tab-store'
 import useEditorTabStore from '~/stores/editor-tab-store'
 import { useNavigationStore } from '~/stores/navigation-store'
 
-export function openInStudio(adapterName: string, filepath: string) {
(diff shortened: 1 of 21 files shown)


src/main/frontend/app/routes/studio/context/element-hover-card.tsx
```diff
@@ -0,0 +1,142 @@
...
+                }}
+              >
+                {value}
+              </span>
+            ))}
+        </div>
+        {element.description && <p className="text-sm" dangerouslySetInnerHTML={{ __html: element.description }} />}
+      </div>
+    </div>,
+    document.body,
```
Matthbo: What does this do?

Daan0709: https://react.dev/reference/react-dom/createPortal if you want to know more 😄

PR #314, opened 2026-03-04
Persist coordinates and dimensions in flow namespace

Daan0709:
- Coordinates are now stored on elements by using `flow:x="xcoord"` and `flow:y="ycoord"`
- Same for width and height `flow:width="width"` and `flow:height="height"`
- Namespace `xmlns:flow="urn:frank-flow"` is automatically added to all configurations that get saved that do not yet have the namespace

afbeelding

### DIFF
--- a/src/test/java/org/frankframework/flow/configuration/ConfigurationControllerTest.java
+++ b/src/test/java/org/frankframework/flow/configuration/ConfigurationControllerTest.java
@@ -103,7 +103,7 @@ class ConfigurationControllerTest {
 		String filepath = "config1.xml";
 		String xmlContent = "<xml>updated</xml>";
 
-		when(configurationService.updateConfiguration(TEST_PROJECT_NAME, filepath, xmlContent))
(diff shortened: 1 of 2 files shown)


src/main/java/org/frankframework/flow/filetree/FileTreeService.java
```diff
@@ -53,8 +54,9 @@ public String readFileContent(String filepath) throws IOException {
         return fileSystemStorage.readFile(filepath);
     }
 
-    public void updateFileContent(String projectName, String filepath, String newContent)
-            throws IOException, ProjectNotFoundException, ConfigurationNotFoundException {
+    public String updateFileContent(String projectName, String filepath, String newContent)
+            throws IOException, ProjectNotFoundException, ConfigurationNotFoundException, Exception,
+                    TransformerException {
```
Matthbo: If you catch Exception by itself there's no need for any child exceptions to be put here.

Because Exception is the parent of all of these it will always give it back no matter what the child exception is

PR #342, opened 2026-03-12
Solved regression bugs

stijnpotters1:
project persistence bugs
configuration saving bugs
file tree context menu bugs
file tree item bugs
configuration tile bugs

Also removed project cache in backend to prevent sync problems. Now the file system is the single source of truth

### DIFF
--- a/src/test/java/org/frankframework/flow/project/ProjectServiceTest.java
+++ b/src/test/java/org/frankframework/flow/project/ProjectServiceTest.java
@@ -657,7 +657,7 @@ public class ProjectServiceTest {
     @Test
     void testToDtoMapsConfigurationFilepaths() throws Exception {
         stubFileSystemForProjectCreation();
-        lenient().when(fileSystemStorage.toRelativePath(anyString())).thenAnswer(inv -> inv.getArgument(0));
+        when(fileSystemStorage.toRelativePath(anyString())).thenAnswer(inv -> inv.getArgument(0));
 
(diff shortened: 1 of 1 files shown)


src/main/frontend/app/routes/studio/canvas/flow.tsx
```diff
@@ -186,6 +190,16 @@ function FlowCanvas() {
     }
   }, [nodes, edges, scheduleAutoSave])
 
+  useEffect(() => {
+    if (!pendingImmediateSave) return
+    useNodeContextStore.getState().setPendingImmediateSave(false)
```
Matthbo: Why set pendingImmediateSave to false before it has actually saved?

src/main/java/org/frankframework/flow/filesystem/CloudFileSystemStorageService.java
```diff
@@ -113,11 +111,14 @@ public Path toAbsolutePath(String path) throws IOException {
...
         if (normalized.startsWith(userRoot)) {
             String relative = normalized.substring(userRoot.length());
-            if (relative.isEmpty()) return "/";
-            if (!relative.startsWith("/")) relative = "/" + relative;
-            return relative;
+
+            while (relative.startsWith("/")) {
+                relative = relative.substring(1);
+            }
+            return relative.isEmpty() ? "" : relative;
```
Matthbo:
```suggestion
            return relative;
```

PR #355, opened 2026-03-17
Implemented array functionality in the datamapper + Fixed extra schematics

VincentvanMedenbach:


src/main/frontend/app/routes/datamapper/property-list.tsx
```diff
@@ -209,29 +212,63 @@ function PropertyList({ config, configDispatch }: PropertyListProperties) {
 
   function openMapping() {
```
stijnpotters1: This method is really long. Perhaps divide this into multiple private helper methods to improve readability

PR #372, opened 2026-03-23
Datamapper - autoscroll when connecting nodes

VincentvanMedenbach:


src/main/frontend/app/hooks/use-datamapper-flow-management.ts
```diff
@@ -70,6 +70,39 @@ export function useFlowManagement({
...
+            x: reactFlowInstance.getViewport().x,
+            y: reactFlowInstance.getViewport().y + 10,
+            zoom: 1, //Don't set this to 0, it results in NaN for X & Y
+          })
+        }
+
+        if (event.clientY > bottomThreshold) {
+          reactFlowInstance.setViewport({
+            x: reactFlowInstance.getViewport().x,
+            y: reactFlowInstance.getViewport().y - 10,
```
stijnpotters1: Here as well

PR #382, opened 2026-03-24
Code formatting while saving

stijnpotters1:
Now the code is not formatted or normalized when the file is getting saved.

Instead of moving elements, attributes and cursor:
Now the cursor stays where it was left and attributes dont shift

Note:
- When adding a configuration or adapter the reformatting is applied.
- Also the full crud for the adapter since this is only used in the studio. When you return to the editor, the code is not formatted without bothering the user while editing the configuration

### DIFF
--- a/src/main/frontend/app/routes/editor/editor.tsx
+++ b/src/main/frontend/app/routes/editor/editor.tsx
@@ -751,7 +751,7 @@ export default function CodeEditor() {
                       applyFlowHighlighter() // Real-time highlight updates
                     }
                   }}
-                  options={{ automaticLayout: true, quickSuggestions: false, tabSize: 3 }}
+                  options={{ automaticLayout: true, quickSuggestions: false, tabSize: 2, insertSpaces: true, detectIndentation: false }}
                 />
               </div>
             </>


Matthbo:
@Daan0709 Could you take a look at this?
I'm not sure if this is a smart idea as the frontend wont represent how the final file will look after saving.
I made this after talking with Niels a bit ago https://github.com/frankframework/flow/issues/403, maybe we should merge that issue into https://github.com/frankframework/flow/issues/380 and update this PR?

Daan0709:
@Matthbo Yeah ill take a look at this. I'll merge #403 into #380 and implement the same file formatting.

Daan0709:
@Matthbo I have merged this branch with master and also added prettier formatting to be in line with what the backend returns. This PR should be ready for review now

src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -159,6 +161,14 @@ function isConfigurationFile(fileExtension: string) {
   return fileExtension === 'xml'
 }
 
+function prettierFormat(xml: string): Promise<string> {
+  return prettier.format(xml, {
+    parser: 'html',
+    plugins: [parserHTML],
+    tabWidth: 3,
+  })
```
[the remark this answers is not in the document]
philipsens: https://github.com/prettier/plugin-xml

src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -422,6 +442,19 @@ export default function CodeEditor() {
...
+
+    editor.addAction({
+      id: 'reformat-xml-prettier',
+      label: 'Reformat',
+      contextMenuGroupId: 'navigation',
+      contextMenuOrder: 3,
+      keybindings: [
+        monacoReference.current.KeyMod.CtrlCmd |
+          monacoReference.current.KeyMod.Shift |
+          monacoReference.current.KeyCode.KeyR,
```
philipsens: A lot of editors use `Alt + Shift + F` or `Crtl + Shift + F`. IntelliJ uses `Ctrl + Alt + L`

I think it would be nice to stick with one of those, for people who have the muscle memory for it already.

src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -422,6 +442,19 @@ export default function CodeEditor() {
         }
       },
     })
+
+    editor.addAction({
+      id: 'reformat-xml-prettier',
+      label: 'Reformat',
+      contextMenuGroupId: 'navigation',
+      contextMenuOrder: 3,
+      keybindings: [
```
philipsens, line 520: The key combination needs to be added to the Help page.

I also saw that there are a lot of other combinations in the editor that are not noted there.

src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -642,7 +685,7 @@ export default function CodeEditor() {
                     scheduleSave()
                     if (value && fileLanguage === 'xml') scheduleSchemaValidation(value)
                   }}
-                  options={{ automaticLayout: true, quickSuggestions: false }}
+                  options={{ automaticLayout: true, quickSuggestions: false, tabSize: 3 }}
```
Matthbo: Tab size of 3?
Do we do this anywhere even in our own code?

Daan0709: I had noticed that our backend returns xml with tabsizes of 3, even though we have the transformer set like so:
```java
transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "2");
```

I had tried a couple different ways of forcing it to be a tab size of 2, but I could not really find a way to make it work. You can reproduce by just creating a new configuration from the configuration manager and then opening that in the editor. 

To keep the format somewhat similar, I opted to just increase the tabsize in the editor instead. But it might be worth it to have someone else take a look at this.

Matthbo: @stijnpotters1 could you see if you can manage to get tabsize of 2 to work for all xml formatting?
And idk if this already works but preferably with spaces instead of tabs

stijnpotters1: @Matthbo Yes i already had in mind that i am going through the project one more time to make sure 2 spaces are applied everywhere in the editor.

stijnpotters1: @Matthbo It was a very easy fix.
Sergi approved, so i will merge it.

PR #384, opened 2026-03-24
Applied custom editor color rules

stijnpotters1:
Changed theme to flow-vs-{theme} instead of vs-{theme} to prevent internal monaco editor theme conflicts

image
image

### DIFF
--- a/src/main/frontend/app/routes/editor/editor.tsx
+++ b/src/main/frontend/app/routes/editor/editor.tsx
@@ -1,10 +1,13 @@
 import RulerCrossPenIcon from '/icons/solar/Ruler Cross Pen.svg?react'
 import Editor, { type Monaco, type OnMount } from '@monaco-editor/react'
+type ITextModel = Monaco['editor']['ITextModel']
+type FindMatch = Monaco['editor']['FindMatch']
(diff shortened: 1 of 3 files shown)


src/main/frontend/package.json
```diff
@@ -26,6 +26,7 @@
     "dagre": "^0.8.5",
     "dotenv": "^17.3.1",
     "isbot": "^5.1.36",
+    "monaco-editor": "^0.55.1",
```
Matthbo: `@monaco-editor/react` brings monaco-editor with it as a dependency, there's no need to add it as a package on its own as well

src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -159,6 +159,76 @@ function isConfigurationFile(fileExtension: string) {
...
+
+  return mapToValidationErrors(flowResult.errors, model).map((err) => ({
+    ...err,
+    lineNumber: err.lineNumber + startLine,
+    startColumn: 1,
+    endColumn: model.getLineLength(err.lineNumber + startLine),
+  }))
+}
+
+async function validateFrank(
```
philipsens: `validateConfiguration`

The term frank is wrongfully used.

PR #413, opened 2026-04-02
Validate nested elements

### LINKED ISSUE #371: Validation on (nested) elements
Currently you can make any/all senders and listeners as nested element in side a receiver for example, but the frankdoc and xsd both have rules that show which nested components are mandatory.
The add nested component button should only appear when nested components are available and element validation should show when an element is invalid because its missing a mandatory nested component.

Daan0709:
- If a node does not have its mandatory children requirements fullfilled, an error block will appear telling the user the node is invalid and what to add to fix it

afbeelding


Matthbo:
So for my visualisation
Does this check the xml itself for missing mandatory attributes or does it check the element objects before its transformed to XML?

Daan0709:
> So for my visualisation Does this check the xml itself for missing mandatory attributes or does it check the element objects before its transformed to XML?



> So for my visualisation Does this check the xml itself for missing mandatory attributes or does it check the element objects before its transformed to XML?

It checks the node object, XML is only generated when saving the flow. So this is all happening before the transformation to XML

Matthbo (approved):

PR #426, opened 2026-04-07
Flow to xml inconsistencies

### LINKED ISSUE #393: Remove forward generation inside Receivers
Image

Currently, everytime a studio flow gets saved, forward elements are generated within the Receiver. These are not allowed to be there

### LINKED ISSUE #425: Exit attributes do not get saved from studio to editor

Daan0709:
- Receivers now no longer generate forwards
- exits persist their attributes

Closes #425, Closes #393


Matthbo (commented):
You now specifically check if an element is a receiver and only add forwards if its not. I was wondering if the Forwards are seen as child elements, if so you should check on that instead to determine when to add Forwards

Daan0709:
> You now specifically check if an element is a receiver and only add forwards if its not. I was wondering if the Forwards are seen as child elements, if so you should check on that instead to determine when to add Forwards

At this point forwards are not yet child elements, instead they are represented by all the edges that are within a flow. Since the receiver is the only element in here that **can** have an edge on the flow but **can not** have a forward in the xml, this seemed like the easiest solution.

Matthbo (approved):

PR #429, opened 2026-04-16
Add modal for configuring non-canvas elements and improve path display

### LINKED ISSUE #397: Implement a configure non-canvas elements screen
Since things like jobs and validators and the like do not belong on the canvas (within a pipeline), they cannot be adjusted in the flow. Therefore a screen should be implemented where users can still set them in the background. This could be done through a button next to the 'Jump to editor' button

stijnpotters1:
image

### DIFF
--- a/src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
+++ b/src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
@@ -1,0 +1,583 @@
+import { createPortal } from 'react-dom'
+import Button from '../inputs/button'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { fetchConfigurationCached, saveConfiguration, clearConfigurationCache } from '~/services/configuration-service'
+import { useProjectStore } from '~/stores/project-store'
+import { showErrorToast } from '~/components/toast'
+import { refreshOpenDiffs } from '~/services/git-service'
+import useEditorTabStore from '~/stores/editor-tab-store'
+import { useFFDoc } from '@frankframework/doc-library-react'
+import { useFrankConfigXsd } from '~/providers/frankconfig-xsd-provider'
+import { parseXsd, getFirstLevelElementsForType } from '~/utils/xsd-utils'
+import ContextInput from '~/routes/studio/context/context-input'
+import type { Attribute, ElementDetails, FFDocJson } from '@frankframework/doc-library-core'
+import { NON_CANVAS_ELEMENTS } from '~/routes/studio/context/palette-config'
+
+interface NonCanvasElement {
+  tagName: string
+  attributeValues: Record<string, string>
+  domChildren: Element[]
+}
+
+interface ElementPickerProps {
+  validTypes: string[]
+  onSelect: (tagName: string) => void
+  onCancel: () => void
+}
+
+interface AttributeEditorProps {
+  element: NonCanvasElement
+  schema: Record<string, Attribute> | undefined
+  ffDoc: FFDocJson | null
+  elements: Record<string, ElementDetails> | null
+  onChange: (updated: NonCanvasElement) => void
+}
+
+interface NonCanvasElementsModalProps {
+  isOpen: boolean
+  onClose: () => void
+  configurationPath: string
+}
+
+const EXCLUDED_TAGS = new Set(['adapter', ...NON_CANVAS_ELEMENTS.map((t) => t.toLowerCase())])
+
+function isExcluded(tagName: string): boolean {
+  return tagName.startsWith('flow:') || EXCLUDED_TAGS.has(tagName.toLowerCase())
+}
+
+function parseNonCanvasElements(xmlString: string): NonCanvasElement[] {
+  const doc = new DOMParser().parseFromString(xmlString, 'text/xml')
+  const result: NonCanvasElement[] = []
+
+  for (const child of doc.documentElement.children) {
+    if (isExcluded(child.tagName)) continue
+
+    const attributeValues: Record<string, string> = {}
+    for (const attribute of child.attributes) {
+      attributeValues[attribute.name] = attribute.value
+    }
+
+    result.push({
+      tagName: child.tagName,
+      attributeValues,
+      domChildren: [...child.children],
+    })
+  }
+
+  return result
+}
+
+function serializeToXml(originalXml: string, elements: NonCanvasElement[]): string {
+  const doc = new DOMParser().parseFromString(originalXml, 'text/xml')
+  const root = doc.documentElement
+
+  for (const child of root.children) {
+    if (!isExcluded(child.tagName)) {
+      child.remove()
+    }
+  }
+
+  const firstAdapter = root.querySelector('Adapter, adapter')
+
+  for (const element of elements) {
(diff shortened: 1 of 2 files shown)


src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
```diff
@@ -0,0 +1,583 @@
...
+
+    for (const domChild of element.domChildren) {
+      newElement.append(doc.importNode(domChild, true))
+    }
+
+    if (firstAdapter) {
+      firstAdapter.before(newElement)
+    } else {
+      root.append(newElement)
+    }
```
Matthbo, line 97: This doesn't really work, the XSD defines the order of elements within a Configuration to have Adapters before any other non-canvas elements.

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
```diff
@@ -0,0 +1,583 @@
...
+      attributeValues,
+      domChildren: [...child.children],
+    })
+  }
+
+  return result
+}
+
+function serializeToXml(originalXml: string, elements: NonCanvasElement[]): string {
+  const doc = new DOMParser().parseFromString(originalXml, 'text/xml')
```
philipsens, line 71: Is it smart to use the DOMParser here? What if some other part of the xml is not wellformed? What parser is used for the flow?

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
```diff
@@ -0,0 +1,583 @@
...
+
+  useEffect(() => {
+    if (!filtered.includes(highlighted)) setHighlighted(filtered[0] ?? '')
+  }, [filtered, highlighted])
+
+  const move = (dir: 1 | -1) => {
+    const idx = filtered.indexOf(highlighted)
+    const next = filtered[Math.max(0, Math.min(filtered.length - 1, idx + dir))]
+    if (next) setHighlighted(next)
+  }
```
philipsens, line 138: I thought move directory. But it is direction. Maybe not use so many abbreviations?

src/main/frontend/app/routes/studio/studio.tsx
```diff
@@ -52,55 +60,80 @@ export default function Studio() {
...
+          onClose={() => setShowNonCanvasModal(false)}
+          configurationPath={activeTabPath}
+        />
+      )}
+      <SidebarLayout name="studio">
+        <>
+          <SidebarHeader side={SidebarSide.LEFT} title="Adapter" />
+          <StudioFileStructure />
+        </>
+        <>
```
Matthbo, line 76: Why are these split up into fragments? feels a little redundant

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
```diff
@@ -0,0 +1,583 @@
...
+                e.preventDefault()
+                move(1)
+                break
+              }
+              case 'ArrowUp': {
+                e.preventDefault()
+                move(-1)
+                break
+              }
+            }
```
philipsens, line 166: The if statements reads a bit hard without the curly brackets.

Also move 1 | -1 could maybe be done with an enum so it reads better `move(direction.UP)`

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
```diff
@@ -0,0 +1,583 @@
...
+
+  const getValue = (key: string) => element.attributeValues[key] ?? ''
+  const setValue = (key: string, value: string) =>
+    onChange({ ...element, attributeValues: { ...element.attributeValues, [key]: value } })
+
+  const extraKeys = Object.keys(element.attributeValues).filter((k) => !allSchemaKeys.includes(k))
+
+  const makeEnumOptions = (attribute: Attribute): Record<string, string> | undefined => {
+    if (attribute.enum && ffDoc?.enums?.[attribute.enum]) {
+      return Object.fromEntries(Object.keys(ffDoc.enums[attribute.enum]).map((k) => [k, k] as [string, string]))
```
philipsens, line 207: Why are you casting the keys? I don't get this

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
```diff
@@ -0,0 +1,583 @@
...
+    }
+  }
+
+  const { mandatory, filled, rest } = useMemo(() => {
+    const mandatory: string[] = []
+    const filled: string[] = []
+    const rest: string[] = []
+
+    for (const key of allSchemaKeys) {
+      const attr = schema![key]
```
philipsens, line 217: Schema can be undefined..?

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
```diff
@@ -0,0 +1,583 @@
...
+        placeholder="e.g. IbisJob"
+        value={value}
+        onChange={(e) => setValue(e.target.value)}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter' && value.trim()) onAdd(value.trim())
+          if (e.key === 'Escape') onCancel()
+        }}
+        className="border-border focus:ring-foreground-active flex-1 rounded border px-2 py-1 text-sm focus:ring focus:outline-none"
+      />
+      <Button onClick={() => value.trim() && onAdd(value.trim())} className="px-2 py-1 text-sm">
```
Matthbo, line 319: Why check for `value.trim()` here when you can also do that in onAdd?
now you're performing the same action twice if its not an empty string

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
Matthbo, line 207: What is happening here?
key to [key, key] and then fromEntries makes no sense

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
philipsens, line 1: I think this is a pretty big file for React components. Is there a real usecase to having all the components in this file, or could it be divided?

philipsens (changes requested):

src/main/frontend/app/components/flow/non-canvas-elements-modal.tsx
Matthbo, line 138: Can't this be done in a native way, HTML tab selection and pressing enter to open?
Honestly this would be a nice accessibility thing but seems a little excessive to have at this point in time

Matthbo (changes requested):
Some non-canvas elements require nested elements to be put into them, now there is no way to do this and only set attributes.
Sergi and I will look at a possible design to improve upon this but a modal like this isn't really going to work.

I'm noticing a lot of variables name that have been shortened into non descriptive names, please leave them descriptive for good readability.

stijnpotters1:
@Matthbo @philipsens Perhaps its better to close this pr for now until you two had the time to make a possible design. Then we can go through the new design with eachother so its clear for me. Then i can give it another go with this pr in the back of my mind.

What do you think?

Matthbo:
Sure, sounds good

PR #431, opened 2026-04-16
Refactor settings components to use the right input components

### LINKED ISSUE #418: Settings inputs
There are special components for in the settings. InputWithLabel for example.
Please go back to an early commit where multiple examples where shown, to make it more consistent.

stijnpotters1:
@Sergi can you verify if this is how you had envisioned it?


philipsens (approved):
Yes i think so

PR #446, opened 2026-04-21
Update error toast styling for improved visibility

### LINKED ISSUE #437: Toast have opacity is too low

stijnpotters1:
Now its back to its original state again
image


Matthbo (approved):
missing-requirements.tsx#40 also uses the `bg-error-background` class. is this also wrong?

stijnpotters1:
> missing-requirements.tsx#40 also uses the `bg-error-background` class. is this also wrong?

@Matthbo Yes this is being used for the missing frank element childs.

image

If we make this background identical to the new toast we cannot read the text inside the missing requirements container due to the contrast changes. So currently it is right.

PR #453, opened 2026-04-23
Refactor editor attribute sorting logic

### LINKED ISSUE #374: Backend automatically alphabetises attributes
Image

Trying to add an attribute can be quite cumbersome because of this, because the attribute will jump when autosaving

stijnpotters1:
image


src/main/frontend/app/routes/studio/flow-to-xml-parser.ts
```diff
@@ -96,6 +107,21 @@ ${pipelineParts.join('\n')}
   </Adapter>`
 }
 
+export function replaceAdapterInXml(configXml: string, adapterIndex: number, newAdapterXml: string): string {
+  const starts = [...configXml.matchAll(/<[Aa]dapter\b/g)].map((m) => m.index)
+
+  if (adapterIndex >= starts.length) return configXml
+
+  const start = starts[adapterIndex]
+  const closeRegex = /<\/[Aa]dapter>/g
```
Matthbo: Instead of doing 2 regex executions for the same thing, is it possible to save the tag name with correct capitalisation so the closing tag can be be something like `</${closingTagName}>`?

Matthbo: The reason is that using regex is discouraged since the javascript regex engine needs to be called in separate instances through the browser's execution of the code. This is understood to be slower than plain string search or replacement and in our case would mean getting the correct name from the first regex (as a capture group result?) and using that for the closing tag

src/main/frontend/app/routes/studio/flow-to-xml-parser.ts
```diff
@@ -167,13 +194,20 @@ function generateXmlElement(
...
-    .join('')}`
-
-  const flowNamespaceString = `flow:y="${roundedY}" flow:x="${roundedX}" flow:width="${width}"${height === undefined ? '' : ` flow:height="${height}"`}`
+  const allAttrs: Record<string, string> = {
+    ...attributes,
+    name,
+    'flow:x': String(roundedX),
+    'flow:y': String(roundedY),
+    'flow:width': String(width),
+    ...(height === undefined ? {} : { 'flow:height': String(height) }),
```
Matthbo: height should be null if not found in node instead of undefined

src/main/java/org/frankframework/flow/utility/XmlFormatter.java
```diff
@@ -0,0 +1,141 @@
...
+		sb.append(" ".repeat(depth * INDENT_SIZE));
+		sb.append("<!--").append(new String(ch, start, length)).append("-->");
+	}
+
+	@Override public void startDTD(String name, String publicId, String systemId) {}
+	@Override public void endDTD() {}
+	@Override public void startEntity(String name) {}
+	@Override public void endEntity(String name) {}
+	@Override public void startCDATA() {}
+	@Override public void endCDATA() {}
```
Matthbo: So many empty overriden methods, are you sure this is the way to go?

Matthbo (changes requested):

philipsens:
Sorry that I'm only jumping in now, but can you explain what order is being used for the attributes now. I commented on the issue (https://github.com/frankframework/flow/issues/374#issuecomment-4321856298), but didn't realize that there was a PR already.

stijnpotters1:
> Sorry that I'm only jumping in now, but can you explain what order is being used for the attributes now. I commented on the issue ([#374 (comment)](https://github.com/frankframework/flow/issues/374#issuecomment-4321856298)), but didn't realize that there was a PR already.

Its sorted in 4 layers.

First on namespaced attributes
Second on mandatory attributes according tot the xsd
Thirth the name attribute since this is used almost in every element.
Fourth the rest of the attributes on alphabetical order

philipsens:
I'd rather see the order that is used in the XSD itself (and required at the top maybe). Then there is "less magic" and is inline with the FF! Doc.

philipsens:
We also discussed with in the Core Refining: 
> Why not just sort in the order of the XSD?
>
> On another note: I don't think the editor should reformat automatically at all. Maybe when > switching context, but never when the user is focused on the editor.

So:

User is typing, saving, editing again -> Nothing should change
User calls reformat function -> Reformat and reorder
User edits flow -> XML gets changes programmatically and gets reformatted and reordered
Maybe: User loses editor focus, changes tab/file or page -> auto-reformat

The order in the FF! Doc is as follows:

* Required attributes
* Order of the XSD

The order of the XSD is actually important because it is sorted by input/output rather than alphabetically. 

The Flow also had namespaces in the XML. I don't care where they are shown (At top or bottom) as long as they are together and also in the order of the XSD.

philipsens (dismissed):
Just to block merging until order and usability have been confirmed/implemented.

stijnpotters1:
> We also discussed with in the Core Refining:
> 
> > Why not just sort in the order of the XSD?
> > On another note: I don't think the editor should reformat automatically at all. Maybe when > switching context, but never when the user is focused on the editor.
> 
> So:
> 
> User is typing, saving, editing again -> Nothing should change User calls reformat function -> Reformat and reorder User edits flow -> XML gets changes programmatically and gets reformatted and reordered Maybe: User loses editor focus, changes tab/file or page -> auto-reformat
> 
> The order in the FF! Doc is as follows:
> 
> * Required attributes
> * Order of the XSD
> 
> The order of the XSD is actually important because it is sorted by input/output rather than alphabetically.
> 
> The Flow also had namespaces in the XML. I don't care where they are shown (At top or bottom) as long as they are together and also in the order of the XSD.

I have implemented this. Now its ordered according to the xsd order.
image

philipsens:
> I have implemented this. Now its ordered according to the xsd order. image

Cool, I think @nielsm5 will be happy with this!

src/main/java/org/frankframework/flow/configuration/ConfigurationService.java
```diff
@@ -26,13 +36,25 @@ public class ConfigurationService {
...
 		this.fileSystemStorage = fileSystemStorage;
 		this.projectService = projectService;
+		this.frankConfigXsdService = frankConfigXsdService;
 		this.fileTreeService = fileTreeService;
 	}
 
+	@PostConstruct
+	public void init() {
+		this.xsdOrderer = loadXsdOrderer();
+	}
```
Matthbo, line 57: Why can't this run at the end in the constructor?

Matthbo (commented):
Dont forget to check the sonar warnings and one comment.
Otherwise it looks good

src/main/java/org/frankframework/flow/configuration/ConfigurationService.java
stijnpotters1: I thought it was more readable to make a PostColnstruct instead of putting the line at the end of the constructor.
I renamed the post construct and added a comment for explanation

Matthbo (approved):

PR #462, opened 2026-05-01
Prevent duplicate <Forward> elements and enhance target node lookup logic

### LINKED ISSUE #460: Every save in the studio results in duplication
```
<ExceptionPipe flow:width="300" flow:x="1264" flow:y="-29" name="Result">
            <Forward name="success" path="Result"/>
            <Forward name="success" path="Result"/>
            <Forward name="success" path="Result"/>
            <Forward name="success" path="eksit"/>
            <Forward name="exception" path="test"/>
```
Every save a
```
<Forward name="success" path="Result"/>
```
keeps being added.

stijnpotters1:


Matthbo:
You now made a seenForwards set in the flow-to-xml parser but isn't this something that should prevented in the glow itself?
Feels like duplicate edges have been created while they shouldn't have, is this something we have control over or do we accept this parser as the place to fix and dedupe such issue?

stijnpotters1:
> You now made a seenForwards set in the flow-to-xml parser but isn't this something that should prevented in the glow itself? Feels like duplicate edges have been created while they shouldn't have, is this something we have control over or do we accept this parser as the place to fix and dedupe such issue?

In the studio you can not use duplicate names anymore when you try to use an element name that is duplicate. For example name="test", the second element will be named test_2 automatically instead of giving an error with: "Name is already used on an element". This is done so the flow stays happy and the user can continue with making a configuration.

Matthbo (approved):

src/main/frontend/app/routes/studio/xml-to-json-parser.ts
```diff
@@ -252,19 +252,27 @@ function addForwardEdges(
   forwardIndexBySourceId: Map<string, number>,
   explicitTargetsBySourceId: Map<string, Set<string>>,
   sourcesWithSuccessExitForward: Set<string>,
-  sourcesWithSuccessPipeForward: Set<string>,
+  sourcesWithSuccessPipeForward: Set<string>
 ) {
   for (const forward of forwards) {
     const targetName = forward.getAttribute('path')
-    if (targetName === null) continue
+    if (!targetName) continue // Veiliger: vangt null en lege strings af
```
stijnpotters1:
```suggestion
    if (!targetName) continue
```

PR #467, opened 2026-05-06
Add glyph for Frank elements and implemented in Editor" functionality

### LINKED ISSUE #447: Add an option to highlight the selected node (listener / pipe) in the code editor
The Flow MVP had this too. But think we can one-up it this time!

It would be really useful to have an option in the node context menu to jump to the adapter in the xml. The option could be called "jump to code", "highlight in xml", "show in xml" or "open in editor". 

It would also be useful to have a shortcut too.

### LINKED ISSUE #448: Add an option to highlight the pipe from within the editor
Like issue #447 but in reverse.

Every pipe would get a glyph that links to the studio and shows the pipe as center point.

I think the nicest options are with a custom glyph, with codelens or with a command.
Glyph: https://microsoft.github.io/monaco-editor/playground.html?source=v0.55.1#example-interacting-with-the-editor-rendering-glyphs-in-the-margin
Codelens: https://microsoft.github.io/monaco-editor/playground.html?source=v0.55.1#example-extending-language-services-codelens-provider-example
Command: https://microsoft.github.io/monaco-editor/playground.html?source=v0.55.1#example-interacting-with-the-editor-adding-a-command-to-an-editor-instance

This one is probably harder to implement than the other issue so the prioraty is a bit lower.

stijnpotters1:
Added glyph and when going from studio to editor the element is highlighted
image
When from editor to studio the element is centered and selected
image

### DIFF
--- a/src/main/frontend/app/routes/editor/editor.tsx
+++ b/src/main/frontend/app/routes/editor/editor.tsx
@@ -517,6 +517,7 @@ export default function CodeEditor() {
   const handleEditorMount: OnMount = (editor, monacoInstance) => {
     editorReference.current = editor
     monacoReference.current = monacoInstance
+    frankGlyphsDecorationsRef.current = null
     setEditorMounted(true)
 
     editor.updateOptions({ glyphMargin: true })
--- a/src/main/frontend/app/routes/editor/editor.tsx
+++ b/src/main/frontend/app/routes/editor/editor.tsx
@@ -841,7 +841,6 @@ export default function CodeEditor() {
                     if (value && fileLanguage === 'xml') {
                       scheduleSchemaValidation(value)
                       applyFlowHighlighter()
-                      applyFrankGlyphs(value)
                     }
                   }}
                   options={{
--- a/src/main/frontend/app/routes/editor/editor.tsx
+++ b/src/main/frontend/app/routes/editor/editor.tsx
@@ -36,10 +36,13 @@ import {
   extractFlowElements,
   findAdapterIndexAtOffset,
   findAdaptersInXml,
+  findElementRangeInXml,
+  findFrankElementsForGlyphs,
   findFlowElementsStartLine,
   lineToOffset,
   wrapFlowXml,
 } from './xml-utils'
+import { openInStudioAtNode } from '~/actions/navigationActions'
 
 type LeftTab = 'files' | 'git'
 type SaveStatus = 'idle' | 'saving' | 'saved'
@@ -255,11 +258,17 @@ export default function CodeEditor() {
   const xsdContentRef = useRef<string | null>(null)
   const errorDecorationsRef = useRef<{ clear: () => void } | null>(null)
   const flowDecorationsRef = useRef<IEditorDecorationsCollection | null>(null)
+  const highlightDecorationsRef = useRef<IEditorDecorationsCollection | null>(null)
+  const frankGlyphsDecorationsRef = useRef<IEditorDecorationsCollection | null>(null)
   const debounceTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const savedTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const validationTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const validationCounterRef = useRef(0)
   const contentCacheRef = useRef<Map<string, CachedFile>>(new Map())
+
+  const [pendingHighlight, setPendingHighlightLocal] = useState<{ subtype: string; name?: string } | null>(
+    () => useEditorTabStore.getState().pendingHighlight,
+  )
 
   const activeTab = useEditorTabStore(
     useShallow((state) => {
@@ -300,6 +309,30 @@ export default function CodeEditor() {
       flowDecorationsRef.current = editor.createDecorationsCollection(decorations)
     }
   }, [fileLanguage])
+
+  const applyFrankGlyphs = useCallback(
+    (content: string) => {
+      const editor = editorReference.current
+      if (!editor || fileLanguage !== 'xml') return
+
+      const elements = findFrankElementsForGlyphs(content)
+
+      const decorations = elements.map((element) => ({
+        range: { startLineNumber: element.startLine, startColumn: 1, endLineNumber: element.startLine, endColumn: 1 },
+        options: {
+          glyphMarginClassName: 'frank-node-glyph',
+          glyphMarginHoverMessage: { value: `Open **${element.name}** in Studio` },
+        },
+      }))
+
+      if (frankGlyphsDecorationsRef.current) {
+        frankGlyphsDecorationsRef.current.set(decorations)
+      } else if (decorations.length > 0) {
+        frankGlyphsDecorationsRef.current = editor.createDecorationsCollection(decorations)
+      }
+    },
+    [fileLanguage],
+  )
 
   const performSave = useCallback(
     (content?: string) => {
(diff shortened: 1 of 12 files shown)


philipsens:
:D Looks nice!

src/main/frontend/app/routes/editor/xml-utils.ts
```diff
@@ -89,3 +90,154 @@ export function findFlowElementsStartLine(xml: string): number {
   }
   return 1
 }
+
+export function findElementInXml(xml: string, subtype: string, name?: string): number | null {
```
philipsens, line 172: I was wondering, since I haven't always been reviewing. How is the parsing of the XML done when generating the nodes? Is done once, or let's say, on every update? The latter was true for the old frank-flow. I don't think that was ideal, but since we did that anyway, I would save the location of each parsed pipe, so when I wend to highlight it, I would already know the location. Would something like that be possible here too? Or is this needed because there is no extra parse taking place?

stijnpotters1: Yes currently parsing does happen repeatedly. I have added a new fix with a cached ref to store the elements that need a glyph

src/main/frontend/app/routes/editor/xml-utils.ts
```diff
@@ -1,24 +1,128 @@
...
+
+function hasNameAttributeWithinTag(lines: string[], startLine: number, targetName: string): boolean {
+  const searchLimit = Math.min(startLine + MAX_LOOKAHEAD_LINES, lines.length)
+
+  for (let i = startLine; i < searchLimit; i++) {
+    if (lines[i].includes(`name="${targetName}"`) || lines[i].includes(`name='${targetName}'`)) {
+      return true
+    }
+
+    if (i > startLine && /^\s*<[A-Za-z]/.test(lines[i])) {
```
philipsens: Maybe add these to the REGEX contants too?

src/main/frontend/app/routes/editor/xml-utils.ts
```diff
@@ -1,24 +1,128 @@
...
 
+export interface FrankElementLocation {
+  subtype: string
+  name: string
+  startLine: number
+  adapterName: string
+  adapterPosition: number
+}
+
+const STRUCTURAL_TAGS = new Set(['Adapter', 'Configuration', 'Module', 'Pipeline', 'Exits'])
```
philipsens: There are some more "Skippable containers":

skippableContainers	
0	"Exits"
1	"Forwards"
2	"Global-forwards"
3	"GlobalForwards"
4	"Module"
5	"PipelinePart"
6	"Root"
7	"Scheduler"

But since that is not a complete list, I'm not sure what to do. Maybe we can make it better in the future. 

Maybe good to make a low-prio issue?

src/main/frontend/app/routes/editor/xml-utils.ts
```diff
@@ -1,24 +1,142 @@
...
+  'GlobalForwards',
+  'PipelinePart',
+  'Root',
+  'Scheduler',
+])
+
+const MAX_LOOKAHEAD_LINES = 15
+const MAX_TAG_LINES = 50
+
+const REGEX_ADAPTER = /<([A-Za-z0-9_:-]+:)?adapter\b[^>]*\bname\s*=\s*["']([^"']*)["']/gi
```
Matthbo, line 31: Why can't the regex be `/<adapter\b[^>]\bname=["']([^"']*)["']/gi`?
Normally there shouldnt be a namespace in front of the Element names right? And the extra whitespace checks seem a little unnecessary too

src/main/frontend/app/routes/editor/xml-utils.ts
```diff
@@ -30,23 +148,125 @@ export function findAdapterIndexAtOffset(adapters: AdapterLocation[], cursorOffs
...
-  const match = xml.match(/<flow:FlowElements[\s\S]*?<\/flow:FlowElements>/)
+  const match = xml.match(REGEX_FLOW_ELEMENTS)
   return match ? match[0] : null
 }
 
 export function wrapFlowXml(fragment: string): string {
-  const inner = fragment
+  const innerContent = fragment
     .replace(/^<flow:FlowElements[^>]*>/, '')
     .replace(/<\/flow:FlowElements>$/, '')
```
Matthbo: This really doesnt have to be regex
```suggestion
    .replace('</flow:FlowElements>', '')
```

src/main/frontend/app/routes/studio/canvas/flow.tsx
```diff
@@ -1062,6 +1131,33 @@ function FlowCanvas() {
     return () => unsubscribe()
   }, [layoutGraph])
 
+  useEffect(() => {
+    function consumePendingSelection() {
```
Matthbo: How many selections are going to be made that it needs a consumer?
In the FF Console I made it so only one highlight was being tracked.
Since the editor will only refer to one element to be highlighted this system seems a little overkill

Matthbo (changes requested):
Also once again variables with single letter names

src/main/frontend/app/routes/editor/xml-utils.ts
stijnpotters1: I thought i was good to prevent users from errors in the flow if they accidentally added an extra whitespace

Matthbo: Well the question is if the extra spaces even remain after saving, or if its formatted out

stijnpotters1: Our XmlFormatterUtils formats attributes by placing them on a new line with indentation padding. If we use [^>] from your suggestion, it expects exactly one character, the regex will fail on our own formatted XML methods. I think i can clean up the reggex more, but i also think we should keep it defensive.

src/main/frontend/app/routes/studio/canvas/flow.tsx
```diff
@@ -994,20 +1058,32 @@ function FlowCanvas() {
...
+        if (pendingSelection) {
+          const tabStore = useTabStore.getState()
+          tabStore.setTabData(tabStore.activeTab, { ...tab, pendingNodeSelection: null })
+
+          requestAnimationFrame(() => {
+            requestAnimationFrame(() => {
+              applySelectionToNodes(pendingSelection)
+              setLoading(false)
+            })
+          })
```
Matthbo: Twice?

src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -636,6 +708,37 @@ export default function CodeEditor() {
     return () => clearTimeout(timeout)
   }, [fileContent, activeTabFilePath, isDiffTab])
 
+  useEffect(() => {
+    return useEditorTabStore.subscribe(
+      (state) => state.pendingHighlight,
+      (highlight) => setPendingHighlightLocal(highlight),
+    )
+  }, [])
```
Matthbo, line 716: Is this supposed to run with every component update?

src/main/frontend/app/routes/studio/canvas/flow.tsx
stijnpotters1: This double frame delay was used to create more time for node positions and selection. But i agree this is not really a good fix. So i have improved it.

src/main/frontend/app/routes/editor/editor.tsx
stijnpotters1: No it runs once on mount of the component

Matthbo: Ah I was confused about the empty dependency list, it sure does look like it

Matthbo (approved):

PR #484, opened 2026-05-13
Add group node color and description support

### LINKED ISSUE #450: Node group improvements
* The group could also have a description.
* The group could have sidebar options
	* Set name
	* Set description
	* Set color
* (maybe) Add the option to remove certain nodes there

stijnpotters1:
image
image

### DIFF
--- a/src/main/frontend/app/app.css
+++ b/src/main/frontend/app/app.css
@@ -60,6 +60,13 @@
     --sticky-color-purple: #e9d5ff;
     --sticky-color-orange: #fed7aa;
 
+    --group-color-blue: #93c5fd;
+    --group-color-violet: #c4b5fd;
+    --group-color-rose: #fda4af;
+    --group-color-green: #86efac;
+    --group-color-amber: #fcd34d;
+    --group-color-cyan: #67e8f9;
+
     /* Palette Styling */
     --palette-pipes: #68d250;
     --palette-listeners: #d250bf;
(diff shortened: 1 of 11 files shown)


philipsens:
What will the pencil in the header of the group do?

stijnpotters1:
> What will the pencil in the header of the group do?

image

@philipsens The name of the group is automatically touched so you can edit the name quick

philipsens:
> @philipsens The name of the group is automatically touched so you can edit the name quick

Ah okay, not bad, but it might be confusing because I thought it might open the sidebar, as edit button. 
We might need to think if we want to do in-place editing for stuff or if we should keep it really strict for now and have all the editing in the sidebar. @Matthbo

Matthbo (commented):
Code looks fine so far, I shall rereview or accept once Sergi is happy

Matthbo:
> > @philipsens The name of the group is automatically touched so you can edit the name quick
> 
> Ah okay, not bad, but it might be confusing because I thought it might open the sidebar, as edit button. We might need to think if we want to do in-place editing for stuff or if we should keep it really strict for now and have all the editing in the sidebar. @Matthbo

I think its better to keep all the editing in the sidebar for consistency.
Having only some things inline and other things not might cause confusion

stijnpotters1:
I have removed the edit possibility to make editing in the right panel more consistent

image

philipsens (approved):
Nice, thank you for implementing Vivy's and my feedback.

PR #485, opened 2026-05-18
Refined studio

### LINKED ISSUE #421: Studio
## General
No file selected/opened and empty configurations text block can be made a lit simpler without borders and shadows.

Open file tabs now always has a shadow on the right border even when not enough files are open to indicate more tabs are on the right.

## Palette
Doc information popup should look more like a mini version of the frankdoc website.
Labels, tags & description should match css with the frankdoc website design (angular components likeness).
Also needs to look a little more wider so everything fits more neatly.
Link to frankdoc should be on the right and a little more filling in the space of the two lines on the left.

Test if having all the categories open by default wouldn't be nicer, maybe this can be configured.

Replace the gradient effect on the elements with a coloured line on the left of the border.

## Canvas
Zooming out of the flow to the point its unreadable should change the elements to become icons instead.

Element border colour isn't passing the vibe in its current state

Ability to copy and paste to another tab (through clipboard so needs to be serialised).

Forwards need more visibility, likely thicker

Size of the element blocks should have a default width and a dynamic height based on the content.
There should be a clear indication that more content exists if the height has been made smaller.

Placement algorithm should be implemented for auto aligning.

Tabs need to show path instead of file.
Open in editor button needs to be similar to Configuration Overview.

stijnpotters1:
- studio open file tabs shadow is now calculated when width changes because of pallete
- palette categories foldout is configurable in the settings now and is also closed on default now
- added centering algorithm
- removed configuration file name at the top and configuration in tab
- reduced min height on nodes
- indicate more contents on nodes where height has been made smaller
- introduced save bar

image


philipsens (changes requested):
The statusbar with the saved text can be moved up again. Otherwise, the tab background does a weird transition to the canvas.

stijnpotters1:
> The statusbar with the saved text can be moved up again. Otherwise, the tab background does a weird transition to the canvas.

Like this?

image

src/main/frontend/app/components/save-status-indicator.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  useEffect(() => {
+    if (!savedAt || saveStatus !== 'saved') return
+    const id = setInterval(() => setTick((t) => t + 1), 30_000)
+    return () => clearInterval(id)
+  }, [savedAt, saveStatus])
+
+  if (saveStatus === 'saving') return <span className="text-foreground-muted text-xs">☁️ Saving...</span>
+  if (saveStatus === 'saved' && savedAt)
+    return <span className="text-foreground-muted text-xs">☁️ Saved {getTimeAgo(savedAt)}</span>
```
philipsens: Use solar icon for cloud

src/main/frontend/app/routes/editor/editor.tsx
```diff
@@ -703,31 +698,19 @@ export default function CodeEditor() {
...
-                    <RulerCrossPenIcon className="h-4 w-4 fill-current" />
-                    Open in Studio
-                  </Button>
-                </div>
+            <div className="flex h-full flex-col">
+              <div className="border-b-border bg-background flex h-10 shrink-0 items-center justify-between border-b px-3">
+                <SaveStatusIndicator />
+                <Button
+                  onClick={handleOpenInStudio}
+                  className="flex items-center gap-1.5 text-xs shadow-sm"
```
philipsens: I don't think this should have a shadow.

philipsens (changes requested):

stijnpotters1:
image

Fixed

src/main/frontend/icons/solar/Cloud.svg
philipsens, line 1: We normally use the outline icon and remove the outer color so we can fill with what we want. Is there a specific reason for using this one?

stijnpotters1: No, i did not know we normally did it this way. I will fix it

philipsens (approved):

PR #487, opened 2026-05-19
Enhance sticky note height calculation for note itself and contents

### LINKED ISSUE #478: Improve notes
- Note in context menu does not take new lines into account
   - Both on small notes and attached node context menu.
- Text in note is not scaling right
- Add note type in context menu, edit not type in note

stijnpotters1:
- Note in context menu does now take new lines into account
     - Both on small notes and attached node context menu.
- Text in note is now auto scaling right in height when typing contents
- Add and edit note will both type in context menu now
- Note indicates more contents when max height is reached

image
image


src/main/frontend/app/stores/flow-store.ts
```diff
@@ -110,6 +110,23 @@ function nextFreeNumericId(nodes: FlowNode[]): number {
   return max + 1
 }
 
+function calculateStickyNoteHeight(text: string, nodeWidth: number): number {
+  const LINE_HEIGHT = 17
+  const VERTICAL_PADDING = 26
+  const HORIZONTAL_PADDING = 24
+  const AVG_CHAR_WIDTH = 6.5
```
Matthbo: Is this based on the font?
Maybe it would be nice to comment the source of where you got this constant, just in case a font change would require a recalculation

src/main/frontend/app/stores/flow-store.ts
```diff
@@ -345,12 +362,12 @@ const useFlowStore = create<FlowState>()(
...
           if (node.id === nodeId && isStickyNote(node)) {
+            const nodeWidth = (node.style?.width as number) ?? node.width ?? FlowConfig.STICKY_NOTE_DEFAULT_WIDTH
+            const newHeight = calculateStickyNoteHeight(text, nodeWidth)
             return {
               ...node,
-              data: {
-                ...node.data,
-                content: text,
-              },
+              style: { ...node.style, width: nodeWidth, height: newHeight },
```
Matthbo: I assume this is required for react flow since it doesn't really work like a DOM element that calculates its own height?

stijnpotters1: Yes thats right!

Matthbo: Gotcha!
This function looks good in that case

src/main/frontend/app/stores/flow-store.ts
stijnpotters1: Good point, i will improve it

philipsens (approved):

PR #491, opened 2026-05-20
Added reveal functionality for active tab in file explorer and enhanced studio shortcuts

### LINKED ISSUE #475: Automatically extend file tree, to where your opened file is located
Maybe add a button here too, to call the action manually when switching tabs a lot.

Image

stijnpotters1:
Click on icon when tab is open and file tree closed
image
file tree opens to this opened tab
image

Also added shortcut actions to the studio file tree
image


src/main/frontend/app/components/file-structure/studio-file-structure.tsx
```diff
@@ -344,20 +409,108 @@ export default function StudioFileStructure() {
...
+                mouseEvent.stopPropagation()
+                triggerItemAction(item.index, studioContextMenu.handleRename)
+              }}
+              onKeyDown={(keyboardEvent) =>
+                keyboardEvent.key === 'Enter' && triggerItemAction(item.index, studioContextMenu.handleRename)
+              }
+            >
+              <Pen className="stroke-foreground h-3.5 w-3.5" />
+            </div>
+          )}
```
philipsens, line 496: These buttons are repeating a lot. I think it would look cleaner if this was a small component. Also because it would be the same in editor-file-tree.

philipsens (commented):
Nice touch disabling the button when it's active.

src/main/frontend/app/components/file-structure/studio-file-structure.tsx
```diff
@@ -310,7 +373,9 @@ export default function StudioFileStructure() {
 
     const pathEndsWithXmlExtension = (item.data as Partial<StudioFolderData>).path?.endsWith('.xml')
 
+    const isRoot = typeof item.data === 'string'
```
Matthbo: `item.data` feels like a strange type here, being able to be both an object with an optional path property, or a string.
pathEndsWithXmlExtension will be `undefined` unless `item.data` is an object with the path property set, I think it would be better to add `?? false` behind it so its always a boolean even if `item.data` is a string or doesn't have a path property.

src/main/frontend/app/components/file-structure/studio-file-structure.tsx
```diff
@@ -100,12 +128,47 @@ export default function StudioFileStructure() {
...
+
+    const itemId = studioTabItemId(activeTab, dataProvider)
+    if (!itemId) return
+
+    const configItemId = itemId.slice(0, itemId.lastIndexOf('/'))
+
+    for (const ancestorId of getAncestorIds(configItemId)) {
+      await dataProvider.loadDirectory(ancestorId)
+      tree.current.expandItem(ancestorId)
+    }
```
Matthbo, line 152: So every part of the directory needs to be loaded separately now?
Doesn't really seem very efficient, unless there's a blocking problem we could have an endpoint that retrieves all the directory data for a full path

src/main/frontend/app/components/file-structure/studio-file-structure.tsx
stijnpotters1: @philipsens Yes i already mentioned this in the draft pr description of #493 last week. Since this pr introduced a new button of this kind i wanted to wait until it was merged so i could rebase on master in #493 and also introduce this button as a component in this pr.

src/main/frontend/app/components/file-structure/studio-file-structure.tsx
stijnpotters1: I have improved it!

src/main/frontend/app/components/file-structure/studio-file-structure.tsx
stijnpotters1: I have improved this problem

src/main/java/org/frankframework/flow/file/FileTreeService.java
```diff
@@ -231,6 +246,35 @@ private FileTreeNode buildShallowTree(Path path, Path relativizeRoot, boolean us
...
+				throw new IllegalArgumentException("Directory does not exist: " + dirPath);
+			}
+
+			boolean useRelativePaths = !fileSystemStorage.isLocalEnvironment();
+			Path relativizeRoot = useRelativePaths ? fileSystemStorage.toAbsolutePath("") : projectPath;
+			return new ProjectDirectory(projectPath, dirPath, relativizeRoot, useRelativePaths);
+		} catch (ApiException _) {
+			throw new IllegalArgumentException("Project does not exist: " + projectName);
+		}
+	}
```
Matthbo, line 269: Git says this is new but doesnt this method exist already?

stijnpotters1: Yes i delegated a big portion of the original getShallowDirectoryTree method into its own method so i could re use it in the new getAncestorPath method

stijnpotters1: Or do you mean something else?

Matthbo: Yes that was it, I see now

Matthbo (approved):

PR #496, opened 2026-05-26
Add folder creation functionality to directory picker

### LINKED ISSUE #434: Make it possible to create folders in the new configuration dialog - or - use the OS file dialog

stijnpotters1:
image
image
image
image


philipsens:
Looks very nice!

Matthbo (commented):
This looks sound in terms of code, is this the correct design to implement though?

stijnpotters1:
> This looks sound in terms of code, is this the correct design to implement though?

What do you mean?

Matthbo:
> > This looks sound in terms of code, is this the correct design to implement though?
> 
> What do you mean?

The border doesn't really seem to fit the design of the page, the buttons feel weird.
What happens if you create a new folder but don't select it, seems to me that creating a new folder means it should be opened at the time of creation instead of having to reselect it

philipsens:
I do kind of agree with Vivy. This is a new design element that isn't getting used in other places. It is better to keep things consistent en use a popup here too, like what a new folder in the editor does.

stijnpotters1:
> I do kind of agree with Vivy. This is a new design element that isn't getting used in other places. It is better to keep things consistent en use a popup here too, like what a new folder in the editor does.

But in this case you get a double popup modal, is this what we want?
It will look like this:
image

philipsens:
> But in this case you get a double popup modal, is this what we want? It will look like this

Ah, yes now it looks the same as in the editor. As a last piece of feedback would be to change 'OK' to 'Create'.

stijnpotters1:
image

philipsens (approved):

PR #497, opened 2026-05-26
Implement file watching functionality in filesystem and directory picker

### LINKED ISSUE #407: Files created outside of the filetree context do not trigger a refresh of the filetree and are not shown
Steps to reproduce: 
1.Open project folder
2.Copy paste a file into the project folder
3.File is not shown on the editor page, 

Image

even if you refresh
If you then proceed to add a file with the editor menu, it does trigger a refresh of the filesystem and the newly copied file is shown

Image

This bug also occurs whenever the datamapper saves its configuration or XSLT

stijnpotters1:
Before adding a file in the original file explorer
image
After adding file in the file explorer its immediately added, even though the directory picker stayed open
image

Also in the editor and studio file tree
image


Matthbo (changes requested):
Lots of undefined's where null's should be used
Making a poller for file changes from the frontend to the backend doesn't seem optimal, websockets or an equivalent should be used to push file changes to the client instead

stijnpotters1:
> Lots of undefined's where null's should be used Making a poller for file changes from the frontend to the backend doesn't seem optimal, websockets or an equivalent should be used to push file changes to the client instead

@Matthbo The current implementation uses SSE (EventSource + Spring's SseEmitter), which is a server push mechanism. The server pushes file change events to the client. It's not polling.

Matthbo:
> > Lots of undefined's where null's should be used Making a poller for file changes from the frontend to the backend doesn't seem optimal, websockets or an equivalent should be used to push file changes to the client instead
> 
> @Matthbo The current implementation uses SSE (EventSource + Spring's SseEmitter), which is a server push mechanism. The server pushes file change events to the client. It's not polling.

Gotcha, seems like I misinterpreted
I've never used EventSource before so this is really awesome to see :D

src/main/java/org/frankframework/flow/file/FileWatcherService.java
Matthbo, line 1: I feel like all SseEmitter specific methods should be in it's own class, it has nothing to do with the file watching itself and then can be reused for other things too

Matthbo (changes requested):
Also notice the sonar warnings

src/test/java/org/frankframework/flow/sse/SseChannelServiceTest.java
Matthbo, line 1: I see 2 new sonar issues here 😬

stijnpotters1: Oops, i did not see those, fixed it!

Matthbo (approved):

PR #509, opened 2026-05-27
Update flow configuration and adjust positioning

### LINKED ISSUE #505: Forward circles are not centered for large adapters
Please test with something like the ZaakBrug, that provides more accurate adapters for developing the Flow.

Image

stijnpotters1:
Zoomed out nodes now also has centered edge dots
image
image


Matthbo:
image
The smaller text seems to be partially hidden when the text gets too long, the background doesn't grow in width when that happens either which now makes it seem like the margin around those components are very arbitrarily decided instead of being calculated.
Is white a good colour for it? Maybe not having a background and border could look nicer

philipsens:
> Is white a good colour for it? Maybe not having a background and border could look nicer

Then the connections might look like there floating.
Maybe an alternative could be, is to position the connection points to the "icon" and place the text above and below it.

We should keep in mind that icons should be used in the future.

I think text should be bigger anyway. 

Here are some options:

image

Matthbo:
I think the left middle one has the most potential, if the forwards are merged towards one point (in the case of many forwards) then this would look very neat

philipsens:
> I think the left middle one has the most potential, if the forwards are merged towards one point (in the case of many forwards) then this would look very neat

Ah yes, so some pipes have many forwards, which might not fit or will go over the text. In the zoomed-out mode, it might be fine to merge everything.

stijnpotters1:
I have improved the styling according to the feedback.
Only the initials have to be changed for icons, but i have made this issue for it #514 
image

philipsens:
So I was wondering, is it still possible to draw a line now?
Also. Why don't the pipes have connection points on the left side? Just to make it look better? I hadn't thought about it. It does look clean, albeit a bit strange maybe.

stijnpotters1:
> So I was wondering, is it still possible to draw a line now? Also. Why don't the pipes have connection points on the left side? Just to make it look better? I hadn't thought about it. It does look clean, albeit a bit strange maybe.

yes it is still possible to draw a line:
image
image

I will apply the other feedback too?

stijnpotters1:
@philipsens i fixed it and now it looks like this:
image

philipsens:
> > So I was wondering, is it still possible to draw a line now? Also. Why don't the pipes have connection points on the left side? Just to make it look better? I hadn't thought about it. It does look clean, albeit a bit strange maybe.
> 
> yes it is still possible to draw a line: image image
> 
> I will apply the other feedback too?

Looks nice! But for the dragging part, you don't know which forward it will get, right? Isn't that totally confusing?

src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
```diff
@@ -115,9 +115,15 @@ export default function FrankNode(properties: NodeProps<FrankNodeType>) {
...
 
-  const compactFirstHandlePosition = useMemo(() => {
-    return (minNodeHeight - (properties.data.sourceHandles.length - 1) * handleSpacing) / 2
-  }, [minNodeHeight, properties.data.sourceHandles.length])
+  const COMPACT_INITIALS_BOX_SIZE = 160
+  const COMPACT_PADDING_TOP = 8
+  const COMPACT_HANDLE_SIZE = 15
+  const COMPACT_HANDLE_GAP = 4
+
+  const compactHandleXOffset = `${(FlowConfig.NODE_DEFAULT_WIDTH - COMPACT_INITIALS_BOX_SIZE) / 2 - COMPACT_HANDLE_SIZE - COMPACT_HANDLE_GAP}px`
```
Matthbo: Maybe this can be its own variable first, same with line 126

stijnpotters1:
> > > So I was wondering, is it still possible to draw a line now? Also. Why don't the pipes have connection points on the left side? Just to make it look better? I hadn't thought about it. It does look clean, albeit a bit strange maybe.
> > 
> > 
> > yes it is still possible to draw a line: image image
> > I will apply the other feedback too?
> 
> Looks nice! But for the dragging part, you don't know which forward it will get, right? Isn't that totally confusing?

image

Matthbo:
> image

Is it me or does the alignment not match between the options and the top text?
The top text also seems to not be vertically centered

src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
```diff
@@ -115,9 +115,19 @@ export default function FrankNode(properties: NodeProps<FrankNodeType>) {
...
+  const COMPACT_HANDLE_SIZE = 15
+  const COMPACT_HANDLE_GAP = 4
+
+  const compactBaseXOffsetPx =
+    (FlowConfig.NODE_DEFAULT_WIDTH - COMPACT_INITIALS_BOX_SIZE) / 2 - COMPACT_HANDLE_SIZE - COMPACT_HANDLE_GAP
+
+  const compactHandleXOffset = `${compactBaseXOffsetPx}px`
+  const compactHandleTop = COMPACT_PADDING_TOP + COMPACT_INITIALS_BOX_SIZE / 2 - COMPACT_HANDLE_SIZE / 2 + 10
+
+  const compactDotTop = compactHandleTop - 8
```
Matthbo: Is the `- 8` correlated with COMPACT_PADDING_TOP?
Would be nice to use that const if it is

Matthbo:
@philipsens other than the text alignment, does it look good to you?

stijnpotters1:
> > image
> 
> Is it me or does the alignment not match between the options and the top text? The top text also seems to not be vertically centered

image
Fixed it

Matthbo (commented):
Looks good to me, let's wait on Sergi's opinion

philipsens:
Selection looks good. I'm not sure how consistent it is with the kebab menu, though.  

I saw this, however:

image

Where first all forwards were grey, they are now colored. I think this is buggy or wrong in this screenshot because there are two forwards and the color indicates success.

stijnpotters1:
I have improved the coloured dots
image

Matthbo (approved):
I think this looks good for now, the positioning can obviously be improved even more with a better algorithm but adding that to this pr would be overkill

PR #512, opened 2026-05-27
Refactor color handling in node components to use theme-based rotating colors

### LINKED ISSUE #502: Wrappers no border?
Image

stijnpotters1:
image

Now all colours in the same way decided as in the frank doc
Now dynamically assigned so Wrapper category colour also fixed


philipsens:
It looks like a lot is blue, doesn't it?

Matthbo:
I agree, Batch, Other & Parameters seem to use the same colour which while might be a coincidence doesn't really seem right
Every element category type does need a unique colour in this case since distinction is important.
The rotating hue trick was a fun idea and works nicely for something like FF Reference but here we need more uniqueness as to not confuse component types with each other

src/main/frontend/app/utils/flow-utils.ts
```diff
@@ -127,22 +128,50 @@ function capitalize(value: string): string {
...
-  const s = 0.9
-  const l = 0.78
-  const a = s * Math.min(l, 1 - l)
-  const f = (n: number) => {
-    const k = (n + h / 30) % 12
-    return l - a * Math.max(-1, Math.min(k - 3, Math.min(9 - k, 1)))
+function hashToCuratedHue(name: string): number {
+  let hash = 5381
+  for (let i = 0; i < name.length; i++) {
+    hash = (hash * 33) ^ (name.codePointAt(i) ?? 0)
```
Matthbo: This part is kinda confusing to me, could you elaborate?
Maybe put a small comment down as well for later reference

philipsens:
> I agree, Batch, Other & Parameters seem to use the same colour which while might be a coincidence doesn't really seem right Every element category type does need a unique colour in this case since distinction is important. The rotating hue trick was a fun idea and works nicely for something like FF Reference but here we need more uniqueness as to not confuse component types with each other

Hmm, it is definitely something to think about. For the Figma design, I did choose some colors, but I hoped that the pastel hue color method would result in nicer - and more - colors. 

We could decide on a middle ground, picking the colors we want ourselves, but generate the color as a fallback for newly added component categories.

src/main/frontend/app/routes/studio/canvas/nodetypes/child-node.tsx
```diff
@@ -174,10 +177,10 @@ export function ChildNodeComponent({
           background: gradientEnabled
             ? `radial-gradient(
               ellipse farthest-corner at 20% 20%,
-              var(--type-${child.type?.toLowerCase()}) 0%,
+              ${getCategoryColor(child.subtype, filters, theme)} 0%,
```
MatthijsSmets: Not sure what the rule of thumb is in react, but in angular its bad practice to call typescript methods from the template for 'static' values.

MatthijsSmets (approved):

PR #516, opened 2026-05-27
Make sure project is singular configuration

### LINKED ISSUE #481: New configuration in overview results in a new src/ nesting

### LINKED ISSUE #501: None singular configurations can be opened as project still
I am not on that location. Only the last folder name is set. Image

Matthbo:
Fixed being able to open directories other than the expected path of a configuration
Also fixes configuration tile names to reflect a better relative path

### DIFF
--- a/src/main/frontend/app/utils/path-utils.ts
+++ b/src/main/frontend/app/utils/path-utils.ts
@@ -1,5 +1,3 @@
-import type { ConfigurationProject } from '~/types/project.types'
-
 /**
  * Extracts the portion of a path after the first occurrence of a marker segment.
  * Returns null if the marker is not found.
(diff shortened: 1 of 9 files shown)


src/main/frontend/app/utils/path-utils.ts
```diff
@@ -9,24 +7,3 @@ export function toRelativePath(absolutePath: string, marker: string): string | n
   const idx = normalized.indexOf(marker)
   return idx === -1 ? null : normalized.slice(idx + marker.length)
```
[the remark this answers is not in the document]
Matthbo: If this path normalisation is used often, I might just make it a util function if it doesn't exist already

src/main/java/org/frankframework/flow/project/ConfigurationProjectService.java
```diff
@@ -126,6 +126,8 @@ public ConfigurationProject openProjectFromDisk(String path) throws IOException,
 		Path absolutePath = fileSystemStorage.toAbsolutePath(path);
 		if (!Files.exists(absolutePath) || !Files.isDirectory(absolutePath)) {
 			throw new ApiException("Project not found at: " + path, HttpStatus.NOT_FOUND);
+		} else if (!absolutePath.endsWith(CONFIGURATIONS_DIR + "/" + absolutePath.getFileName())) {
```
philipsens: What if the path is different? With all the Docker and Kubernetes directories, they get mounted and don't always have the same name.

We could also check for one of the following files in the root of the dir. `Configuration.xml` or maybe `(!BuildInfo)*.properties`

src/main/java/org/frankframework/flow/project/ConfigurationProjectService.java
```diff
@@ -126,6 +126,8 @@ public ConfigurationProject openProjectFromDisk(String path) throws IOException,
...
+			throw new ApiException("Provided path doesn't seem to be a singular configuration", HttpStatus.BAD_REQUEST);
```
philipsens: Is this an error the user could see? Because then I would change the wording a bit.

src/main/java/org/frankframework/flow/project/ConfigurationProjectService.java
Matthbo: Hmm this is giving me ideas, currently even for new projects a certain folder structure is expected, I just changed what was already implemented when factoring from full project to per configuration setup but this whole folder structure enforcement is wrong.

Your suggestion should also be implemented for all the other ways of creating / opening a configuration.

Matthbo:
image

Updated the styling a little too

philipsens (approved):

PR #518, opened 2026-05-29
Add onSelect prop to ChildNode component for selection handling

### LINKED ISSUE #503: Nested elements need two clicks
Feels weird to select the parent first and then double click for nested, but then clicking again selects the parent. I think one click should be enough. Clicking outside the nested element should select the parent, the area is large enough. Also, there should probably be an indication that the nested element is selected, now the parent has the highlight.

stijnpotters1:


src/main/frontend/app/routes/studio/canvas/nodetypes/child-node.tsx
```diff
@@ -158,12 +171,19 @@ export function ChildNodeComponent({
   return (
     <div
       data-childnode-id={child.id}
-      className={`bg-background relative mr-0.5 mb-2 rounded-md border ${dragForbidden ? 'border-2 border-dashed' : 'border-border'}`}
+      className={`bg-background relative mr-0.5 mb-2 rounded-md border ${isSelected ? 'border-1' : dragForbidden ? 'border-2 border-dashed' : 'border-border'}`}
```
philipsens: This would probably be better with clsx and some and operators.

philipsens (changes requested):

src/main/frontend/app/routes/studio/canvas/nodetypes/child-node.tsx
stijnpotters1: Yeah i just saw the linting nesting error. I am going to fix it!

src/main/frontend/app/routes/studio/canvas/flow.tsx
Matthbo: Unused?

Matthbo (commented):
I see some build warnings about hooks with missing dependencies, is this correct?

src/main/frontend/app/routes/studio/canvas/flow.tsx
stijnpotters1: Oops i forgot to push this removed piece

philipsens (approved):

PR #534, opened 2026-06-02
Refactored SSE watcher to manage multiple event sources

### LINKED ISSUE #521: Improve watch efficiency in frontend
Open watch at beginning and close at end, instead of opening and closing on every routing change

stijnpotters1:
Before every component was calling useSseWatcher with the same URL would open its own EventSource connection. When the component unmounted, it would close it immediately and this continued.

Now if two components watch the same URL, they reuse the same EventSource. Each registers its own handler on that shared connection. The connection only closes when the last subscriber unmounts

### DIFF
--- a/src/main/frontend/app/hooks/use-file-watcher.ts
+++ b/src/main/frontend/app/hooks/use-file-watcher.ts
@@ -1,6 +1,14 @@
 import { useEffect, useRef } from 'react'
 import { apiUrl } from '~/utils/api'
 
+interface WatcherEntry {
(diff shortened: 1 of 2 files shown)


Matthbo (commented):
Wow that's easy
The only thing that could be improved is to keep the watcher open as long as the project is open. Now it seems to close on the configuration overview page and reopen once navigated back to either the editor or studio page.
Is this possible?

Matthbo (approved):
Great 😄

PR #536, opened 2026-06-02
Refactor color handling and improve theme support in components

philipsens:
Same as the other PR, but this time we get to review it. 🥲


stijnpotters1:
@philipsens Oops..

Matthbo:
> We could decide on a middle ground, picking the colors we want ourselves, but generate the color as a fallback for newly added component categories.

I think this is something we have to implement before assuming the colour roulette will work out.
This way new categories get assigned a neat colour that shouldn't clash while making sure all existing categories are unique

Issue #507, opened 2026-05-26
Super duper Studio canvas update


philipsens:
The Studio canvas has some work left to be done to match the Figma design I did a year ago. Some functionalities in the design are really helping with UX, like circle color and line type.

Items are better alighted than now and attributes have ellipses when to long. 

<img width="6615" height="6009" alt="Image" src="https://github.com/user-attachments/assets/61254312-284b-4334-998a-9c0f8bba7bf2" />

The forwards were hard to draw. So ignore the line shape (and missing label)

_For a better overview, sub-issues need to be created_

philipsens:
# Forwards

## Label position

### Problem

Labels on long forwards are difficult to read and require a lot of pan-zooming.

### Solution

Add the label just a bit after and before the connection circles on both sides. 
If possible, only one in the middle if the forward is really short.

## Connection circles 

### Problem

In zoomed-out mode, the circles are difficult to see. And when merged, it is not possible to see the difference.

### Solution

Add the colors to the labels. When wanting to drag a new forward, the dropdown can be used.

PR #537, opened 2026-06-02
Enhanced node components with improved shadow effects for better visual depth

### LINKED ISSUE #523: Frank node parent child shadows
box shadow
inline shadow

om en om

stijnpotters1:
Before:
image

After:
According to the #507 design
image

Global view:
image


philipsens:
I think the shadow should go down. And the nested element has a weird inset shadow.

stijnpotters1:
> I think the shadow should go down. And the nested element has a weird inset shadow.

I have improved it
image
What so you think?

philipsens:
image

Shadown only has offset down

stijnpotters1:
Like this?
image

philipsens:
When looking at the code, you did a pretty manual job for something that is built into Tailwind CSS. https://tailwindcss.com/docs/box-shadow

Just use shadow-md and inset-shadow-md. 

I noticed that the configuration page also had some shadows and inset shadows. Could you make the shadows consistent over these two places?

stijnpotters1:
> When looking at the code, you did a pretty manual job for something that is built into Tailwind CSS. https://tailwindcss.com/docs/box-shadow
> 
> Just use shadow-md and inset-shadow-md.
> 
> I noticed that the configuration page also had some shadows and inset shadows. Could you make the shadows consistent over these two places?

image

philipsens:
I think nested components should somewhat look like the parent component. Also having a container with shadow for thier nested elements. It might be possible to have the same react component for it, minus the forwards.

stijnpotters1:
image

I have made components to unify the styling of parent and childs.
Also improved the shadows slightly

philipsens:
> image
> 
> I have made components to unify the styling of parent and childs. Also improved the shadows slightly

Param should not be able to be set on the SenderPipe??

Also, I think it is alright.

stijnpotters1:
> > image
> > I have made components to unify the styling of parent and childs. Also improved the shadows slightly
> 
> Param should not be able to be set on the SenderPipe??
> 
> Also, I think it is alright.

Okay i have made an additional issue for the params in the sender. So you can approve it if you want

philipsens (approved):

PR #542, opened 2026-06-03
Remove editor settings from settings menu items for future implementation

### LINKED ISSUE #525: No editor settings results in error
Image

Perhaps just hide for now like the datamapper tab?

stijnpotters1:


src/main/frontend/app/routes/settings/settings-menu-items.ts
```diff
@@ -51,14 +58,14 @@ const SettingsMenuItems = {
...
-    },
-  },
+  // editor: {
+  //   index: 'editor',
+  //   data: {
+  //     title: 'Editor',
+  //     description: 'Editor settings',
+  //     icon: CodeIcon,
+  //   },
+  // },
```
Matthbo:
```suggestion
  /*editor: {
    index: 'editor',
    data: {
      title: 'Editor',
      description: 'Editor settings',
      icon: CodeIcon,
    },
  },*/
```

philipsens: Any linter rules on this?

Matthbo: Maybe we should configure this rule in all our projects?
https://eslint.org/docs/latest/rules/multiline-comment-style

philipsens: New one: https://eslint.style/rules/multiline-comment-style

Actually, we shouldn't allow commented code at all. Only comments for docs.

src/main/frontend/app/routes/settings/settings-menu-items.ts
Matthbo, line 1: Last time Sergi asked why these comments even need to be here, its not that hard to add the data by looking at how the other pages are done, the extra todo comment above it is even really extra here and not needed

stijnpotters1: I thought it would be handy to keep it commented since this probabily will be reintroduced when its necessary. But i will remove it then

Matthbo (approved):

PR #549, opened 2026-06-15
Solved Cookie token creation for requests

stijnpotters1:
Server (CsrfCookieFilter)  →  Set-Cookie: XSRF-TOKEN  →  browser stores it
 browser (api.ts)   →  reads cookie, sets header  →  X-XSRF-TOKEN on POST
 Server (CsrfFilter)   →  header == cookie?   →  yes → allow

 The reason the filter was needed is that without it the server never issued the cookie in the first place, so the frontend had nothing to read and copy onto the request. The filter supplies the token (via a response cookie) and the frontend is what actually attaches it to requests.

This also directly solved issue #548

PR #550, opened 2026-06-15
Implemented uniformal xml insertion when creating an adapter or configuration file

### LINKED ISSUE #489: When adding a new xml file, add sample adapter

stijnpotters1:
first merge #549


Matthbo (commented):
2 things

I see that creating a new adapter now also returns the xmlcontent to the frontend, why is this done like this instead of returning a http 204 and loading the content once the file is opened?
You now look through the adapters inside the xml to open the file in the studio but isnt there existing functionality for this? Or maybe creating a new file doenst immedately have to open it in the studio but instead open the studio file browser on the right folder so the user can choose

I see a lot of `replace('\\', '/')`, maybe we should always do this to paths coming from the filesystem so we dont have to do this everywhere all the time in both back and frontend

src/main/java/org/frankframework/flow/adapter/AdapterController.java
```diff
@@ -44,9 +45,9 @@ public ResponseEntity<Void> updateAdapter(@RequestBody AdapterUpdateDTO dto) thr
 	}
 
 	@PostMapping("/{projectName}/adapters")
-	public ResponseEntity<Void> createAdapter(@PathVariable String projectName, @RequestBody AdapterCreateDTO dto) throws IOException {
-		adapterService.createAdapter(dto.configurationPath(), dto.adapterName());
-		return ResponseEntity.ok().build();
+	public ResponseEntity<AdapterLocationDTO> createAdapter(@PathVariable String projectName, @RequestBody AdapterCreateDTO dto) throws IOException {
+		int adapterPosition = adapterService.createAdapter(dto.configurationPath(), dto.adapterName());
```
Matthbo, line 49: Why does the frontend need to get the adapter position?

src/main/frontend/app/components/file-structure/use-file-tree-context-menu.ts
```diff
@@ -127,9 +133,20 @@ export function useFileTreeContextMenu({
             return
           }
 
+          const filePath = `${parentPath}/${name}`
+          const isXml = name.toLowerCase().endsWith('.xml')
+          const configsRoot = configurationsRootPath ? normalizePath(configurationsRootPath) : undefined
```
Matthbo: null instead of undefined when initialising a variable

Matthbo (commented):
I still see a few strange places in the frontend where normalizePath() is used, are you sure the paths can't be normalized before getting to those points?

Other than that a couple of small things but looks good so far!

src/main/java/org/frankframework/flow/adapter/AdapterController.java
stijnpotters1: The studio identifies an adapter by its index within the config file, not only its name.  Because a config can hold several adapters and we don't enforce unique names the tab id is filepath::adapterName::position and the canvas resolves it via allAdapters[position]. And we fetch the adapter after creating it through this position

Matthbo (approved):

PR #554, opened 2026-06-17
Enhance configuration file handling, validate filename input and adjust paths

### LINKED ISSUE #544: New Configuration file name / path shouldn't end in \ or /
Having a path end with `\` or `/` breaks the loaded configuration somehow, Configuration Overview shows no configuration is loaded

stijnpotters1:
Configuration overview now shows default Configuration.xml file that is created on default after creating a new configuration.
image

Added validation so you cannot end file or path with '/'
image
image

### DIFF
--- a/src/main/frontend/app/routes/projectlanding/clone-configuration-modal.tsx
+++ b/src/main/frontend/app/routes/projectlanding/clone-configuration-modal.tsx
@@ -4,6 +4,7 @@ import Button from '~/components/inputs/button'
 import CloseButton from '~/components/inputs/close-button'
 import Input from '~/components/inputs/input'
 import { filesystemService } from '~/services/filesystem-service'
+import { joinPath } from '~/utils/path-utils'
 
 type CloneProjectModalProperties = {
   isLocal: boolean
@@ -39,6 +40,9 @@ export default function CloneConfigurationModal({
     .pop()
     ?.replace(/\.git$/i, '')
 
+  const targetName = repoName || 'cloned-project'
+  const targetPath = location ? joinPath(location, targetName) : targetName
+
   const handleClone = () => {
     if (!repoUrl.trim()) return
     if (isLocal && !location) return
@@ -53,7 +57,7 @@ export default function CloneConfigurationModal({
       finalPath = location ? `${location}/${name}` : name
     }
 
-    onClone(repoUrl.trim(), finalPath, token)
+    onClone(repoUrl.trim(), finalPath, token || undefined)
     handleClose()
   }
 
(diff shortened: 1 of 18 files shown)


Matthbo (commented):
Let's discuss this PR next time we're both available

src/main/java/org/frankframework/flow/configuration/ConfigurationService.java
```diff
@@ -97,15 +95,10 @@ public String updateConfiguration(String projectName, String filepath, String co
 
 	public String addConfiguration(String projectName, String configurationName) throws IOException, ApiException, TransformerException, ParserConfigurationException, SAXException {
 		ConfigurationProject configurationProject = configurationProjectService.getProject(projectName);
-		Path absProjectPath = fileSystemStorage.toAbsolutePath(configurationProject.getRootPath());
-		Path configDir = absProjectPath.resolve(CONFIGURATIONS_DIR).normalize();
```
Matthbo: Why remove the 'src/main/configurations' base path?

stijnpotters1: Because in the flow a configuration is opened which is already inside the src/main/configurations directory. So we do not have to set the additional src/main/configurations folder structure inside the configuration to add another configuration file.

Matthbo: Ah so this has nothing to do with creating or opening a configuration?
We do want to enforce projects to put their configuration in `<root>/src/main/configurations/<configuration name>/configuration.xml` or later whatever the default name is for `configuration.xml`

stijnpotters1: No it is for adding a file to a configuration, not the configuration itself

Matthbo: The method is called `addConfiguration` which in this case would be a little unfortunate, `addConfigurationFile` would make more sense

src/main/frontend/app/routes/projectlanding/clone-configuration-modal.tsx
```diff
@@ -53,7 +57,7 @@ export default function CloneConfigurationModal({
       finalPath = location ? `${location}/${name}` : name
     }
 
-    onClone(repoUrl.trim(), finalPath, token)
+    onClone(repoUrl.trim(), finalPath, token || undefined)
```
Matthbo: this `|| undefined` really doesnt matter since empty strings will be checked later on

Matthbo (approved):
Small comment, otherwise approved

PR #558, opened 2026-06-17
Add flow namespace repair functionality and Modules can have root flow namespace now

### LINKED ISSUE #556: Opening an adapter in the studio removes adapter from configuration
Perhaps something gets deleted because of a restructure after opening the file?
Which results in a not recognized adapter

Note:
Only happens when opening a zaakbrug configuration.
Does not happen with a configuration that is made within the flow itself.

Before opening
Image
After opening:
Image

stijnpotters1:
- Modules now also are configured as possible root elements and also can have flow namespaced elements
- Broken configurations are auto repaired if they have this problem now.

image

### DIFF
--- a/src/main/java/org/frankframework/flow/configuration/ConfigurationService.java
+++ b/src/main/java/org/frankframework/flow/configuration/ConfigurationService.java
@@ -124,7 +124,7 @@ public class ConfigurationService {
 			return xml;
 		}
 
-		return xml.replaceFirst("(<(?:Configuration|Module))\\b", "$1 xmlns:flow=\"urn:frank-flow\"");
+		return XmlConfigurationUtils.addFlowNamespaceDeclaration(xml);
 	}
 
 	private String loadDefaultConfigurationXml() throws IOException {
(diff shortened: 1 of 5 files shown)


src/main/java/org/frankframework/flow/configuration/ConfigurationService.java
```diff
@@ -123,7 +124,7 @@ private String ensureFlowNamespace(String xml) {
 			return xml;
 		}
 
-		return xml.replaceFirst("(<Configuration)\\b", "$1 xmlns:flow=\"urn:frank-flow\"");
+		return xml.replaceFirst("(<(?:Configuration|Module))\\b", "$1 xmlns:flow=\"urn:frank-flow\"");
```
Matthbo: For the future, shouldn't we make this a list? or somehow even fetch it from frankdoc (if possible)?

src/main/java/org/frankframework/flow/utility/XmlConfigurationUtils.java
```diff
@@ -45,6 +45,16 @@ public static Document insertFlowNamespace(String configurationXml)
...
 	}
 
+	public static String repairFlowNamespace(String configurationXml) {
+		if (configurationXml == null
+				|| configurationXml.contains("xmlns:flow")
+				|| !configurationXml.contains("flow:")) {
+			return configurationXml;
+		}
+
+		return configurationXml.replaceFirst("(<(?:Configuration|Module))\\b", "$1 xmlns:flow=\"urn:frank-flow\"");
```
Matthbo: Yeah definitely at least a reusable list, see first comment for more info

Matthbo (approved):

PR #559, opened 2026-06-17
Refactor project import functionality to support zip file uploads and enforce size limits

### LINKED ISSUE #482: Can not import configuration folder anymore

stijnpotters1:
image
image
image
image

With a file too big:
image

### DIFF
--- a/src/main/frontend/app/services/project-service.ts
+++ b/src/main/frontend/app/services/project-service.ts
@@ -4,13 +4,11 @@ import type { ConfigurationProject } from '~/types/project.types'
 
 /**
  * Upper bound for the zipped import, kept in sync with the nginx `client_max_body_size`.
- * The check runs on the compressed archive, so the original folder may be considerably larger.
  */
 export const MAX_IMPORT_ZIP_BYTES = 80 * 1024 * 1024
 
 /**
  * Upper bound for the *uncompressed* project, mirrored from the backend import guard.
- * Checked before uploading so an oversized folder fails fast without a round-trip.
  */
 export const MAX_IMPORT_UNCOMPRESSED_BYTES = 80 * 1024 * 1024
 
@@ -79,8 +77,7 @@ export async function importProjectFolder(files: FileList): Promise<Configuratio
   const projectName = files[0].webkitRelativePath.split('/')[0]
 
   /*
-   * Cheap pre-check on the uncompressed size so an oversized folder fails fast,
-   * before we spend time reading and zipping it. Mirrors the backend's 413 guard.
+   * pre-check on the uncompressed size so an oversized folder fails fast,
    */
   let uncompressedBytes = 0
   for (const file of files) {
--- a/src/main/frontend/app/services/project-service.ts
+++ b/src/main/frontend/app/services/project-service.ts
@@ -1,4 +1,4 @@
-import { zipSync } from 'fflate'
+import { zip } from 'fflate'
 import { apiFetch, apiUrl } from '~/utils/api'
 import type { ConfigurationProject } from '~/types/project.types'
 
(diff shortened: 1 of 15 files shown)


Matthbo (commented):
The max uncompressed size is 80MB, but what if you'd want to increase this as a non-developer?
Maybe it should be an app constant so its configurable for the backend and the frontend can request the max upload restrictions from the api

stijnpotters1:
> The max uncompressed size is 80MB, but what if you'd want to increase this as a non-developer? Maybe it should be an app constant so its configurable for the backend and the frontend can request the max upload restrictions from the api

As a non-developer there is no way to increase the 80MB size. Even if we implement a configurable constant in the app that provides the possibility to increase the size, the nginx client_max_body_size will prevent the folder to even end up in the app. This is a global constant in nginx and we can not dynamically change this per client or session

Matthbo:
> eloper there is no way to increase the 80MB size. Even if we implement a configurable constant in the app that provides the possibility to increase the size, the nginx client_max_body_size will prevent the folder to even end up in the app. This is a global constant in nginx and we can not dynamically change this per client or session

No but a system admin usually can, not even giving the option means we might walk into this issue sooner or later, giving the system administrator the power to configure it means it will be their problem instead of a technical issue

src/main/frontend/app/services/project-service.ts
```diff
@@ -52,21 +65,51 @@ export async function exportProject(projectName: string): Promise<void> {
...
 }
+
+function zipAsync(entries: Record<string, Uint8Array>): Promise<Uint8Array<ArrayBuffer>> {
+  return new Promise((resolve, reject) => {
+    zip(entries, { level: 6 }, (error, data) => {
+      if (error) reject(error)
+      else resolve(data as Uint8Array<ArrayBuffer>)
+    })
+  })
+}
```
Matthbo, line 115: This could be an async function to get the reject behaviour you throw the error

Matthbo (changes requested):

src/main/frontend/app/services/project-service.ts
Matthbo: Oh it didnt grab the correct lines, sorry
But it is that whole function since it now just return new promise which an async function does automatically

Matthbo: So now you made it sync instead of async, does that work in terms of performance, wont this block the browser from responding till its done?
The function name now doesn't make much sense either

I more so thought this couldve been a nice solution
```
function zipAsync(entries: Record<string, Uint8Array>): Promise<Uint8Array<ArrayBuffer>> {
  return zip(entries, { level: 6 })
}
```
or this because typescript isn't the brightest
```
async function zipAsync(entries: Record<string, Uint8Array>): Promise<Uint8Array<ArrayBuffer>> {
  try {
    return await zip(entries, { level: 6 })
  } catch (error) {
    // the try-catch is only so the error wont get voided, blame ts for not liking the first example
    throw error
  }
}
```

Matthbo (changes requested):

src/main/frontend/app/services/project-service.ts
stijnpotters1: I have combined your option and mine, because your solution did not return the right type. Now it works good and is awaitable.

stijnpotters1:
```
async function zipAsync(entries: Record<string, Uint8Array>): Promise<Uint8Array<ArrayBuffer>> {
  try {
    return await new Promise<Uint8Array<ArrayBuffer>>((resolve, reject) => {
      zip(entries, { level: 6 }, (error, data) => {
        if (error) reject(error)
        else resolve(data as Uint8Array<ArrayBuffer>)
      })
    })
  } catch (error) {
    // the try-catch is only so the error wont get voided
    throw error
  }
}
```

for this solution i see that the try catch is redundant because the async function already propagates the rejection. so i will convert it to:

```
async function zipAsync(entries: Record<string, Uint8Array>): Promise<Uint8Array<ArrayBuffer>> {
  return await new Promise<Uint8Array<ArrayBuffer>>((resolve, reject) => {
    zip(entries, { level: 6 }, (error, data) => {
      if (error) reject(error)
      else resolve(data as Uint8Array<ArrayBuffer>)
    })
  })
}
```

Matthbo (approved):

PR #569, opened 2026-06-23
Refactor interfaces to types for improved consistency across the code base

### LINKED ISSUE #370: Prefer types over interfaces in TS files
Add eslint rule `'@typescript-eslint/consistent-type-definitions': ['error', 'type']` and replace `interface` with `type` equivalents.

The reason for this is that types cant override each other while interfaces will without error.
If 2 interfaces exist with the same name, doesn't matter if they are in different modules, one will override the other.
Types don't do this, types handled per module and can't have duplicates

stijnpotters1:


src/main/frontend/app/types/global.d.ts
```diff
@@ -1,3 +1,5 @@
-﻿interface Window {
+﻿// Must stay an `interface` to augment the global `Window` (declaration merging).
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+interface Window {
   showDirectoryPicker(): Promise<FileSystemDirectoryHandle>
```
Matthbo: Im a little curious, did this have to live on the window or may there be another way to do this, like through a service or something
That would have my preference

stijnpotters1: I just found out this is a dead piece of code. So i can just delete it

Matthbo (approved):

PR #574, opened 2026-06-24
Add non-canvas element management with controller and service integration

### LINKED ISSUE #397: Implement a configure non-canvas elements screen

### LINKED ISSUE #600: Resize border is blue instead of FF yellow
Image

stijnpotters1:
image
image
image
image
image


philipsens:
Looks good! 

There is a difference between the pen-pot tho. In the design, there can be two rows in each component. A type and a name (or ref in the case of an include). The type should be small and bold while the name would be larger. Is there is no name, the type should be the same styling as the name would've been.

Also, I think elements should just be called components.

image

Also, I think each component should be clickable to configure it, even if it has a button to go to the studio. That makes it easy to rename an adapter for example or to delete it.

stijnpotters1:
image

Improved!

src/main/frontend/app/routes/configurations/configuration-file-tile.tsx
```diff
@@ -1,29 +1,111 @@
...
+  const handleDragEnter = (event: DragEvent<HTMLDivElement>) => {
+    if (!isComponentDrag) return
+    event.preventDefault()
+    dragDepth.current += 1
+    setIsDropTarget(true)
+  }
+
+  const handleDragLeave = () => {
+    if (!isComponentDrag) return
+    dragDepth.current -= 1
```
Matthbo, line 70: No preventDefault() needed here?

src/main/frontend/app/routes/configurations/configuration-file-tile.tsx
Matthbo, line 1: Maybe sub components defined here should be in their own files

src/main/frontend/app/routes/configurations/configuration-overview.tsx
```diff
@@ -134,27 +219,51 @@ export default function ConfigurationOverview() {
...
+  const allConfigFiles = useMemo((): ConfigurationFile[] => {
+    const files = configFiles.map((file) => ({
+      path: file.path,
+      relativePath: file.relativePath,
+      adapterNames: file.adapterNames ?? [],
+    }))
+
+    return files.toSorted((a, b) => {
+      const aIsRoot = a.relativePath.split(/[/\\]/).pop()?.toLowerCase() === 'configuration.xml'
+      const bIsRoot = b.relativePath.split(/[/\\]/).pop()?.toLowerCase() === 'configuration.xml'
```
Matthbo: Are these paths not normalised at this point?
If so then you wont have to check for both `/` and `\` all the time

src/main/java/org/frankframework/flow/noncanvascomponent/NonCanvasComponentService.java
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	private void writeConfigurationDocument(String configurationPath, Document configurationDocument) {
+		Path absolutePath = resolveExistingConfiguration(configurationPath);
+
+		try {
+			String updatedContent = XmlConfigurationUtils.convertNodeToString(configurationDocument);
+			fileSystemStorage.writeFile(absolutePath.toString(), updatedContent);
+		} catch (TransformerException | IOException exception) {
+			throw new ApiException("Failed to write configuration: " + exception.getMessage(), HttpStatus.BAD_REQUEST);
```
Matthbo: How is this a bed request? Something went wrong on the server, not because of the client

src/main/java/org/frankframework/flow/noncanvascomponent/NonCanvasComponentService.java
```diff
@@ -0,0 +1,135 @@
...
+
+	private Document readConfigurationDocument(String configurationPath) {
+		Path absolutePath = resolveExistingConfiguration(configurationPath);
+
+		try {
+			String content = fileSystemStorage.readFile(absolutePath.toString());
+			String repairedContent = XmlConfigurationUtils.repairFlowNamespace(content);
+			return XmlSecurityUtils.createSecureDocumentBuilder().parse(new InputSource(new StringReader(repairedContent)));
+		} catch (IOException | ParserConfigurationException | SAXException exception) {
+			throw new ApiException("Failed to read configuration: " + exception.getMessage(), HttpStatus.BAD_REQUEST);
```
Matthbo: Same here, if the file is not found then yes its on the client, but any other error is the fault on the server side

src/main/frontend/app/services/non-canvas-component-service.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    for (const childName of childNames) collect(childName)
+  }
+
+  for (const rootType of ROOT_TYPE_CANDIDATES) {
+    for (const name of getFirstLevelElementsForType(xsdDocument, rootType)) collect(name)
+  }
+
+  return [...addableNames].toSorted((first, second) => first.localeCompare(second))
```
Matthbo: toSorted already makes a new array copy, no need to spread onto a new array first

src/main/frontend/app/routes/configurations/non-canvas-component-context.tsx
```diff
@@ -0,0 +1,245 @@
...
+          >
+            {isSaving ? 'Saving...' : 'Save & Close'}
+          </Button>
+
+          <Button className="w-auto" onClick={mode === 'edit' ? handleDelete : onClose} disabled={isSaving}>
+            Delete
+          </Button>
+        </div>
+
+        {errorMessage && <p className="text-error mt-2 text-sm">{errorMessage}</p>}
```
Matthbo: This is similar to how its done in the studio right?
Cant parts of that logic be its own component and reused here?

Matthbo (changes requested):

src/main/frontend/app/routes/configurations/configuration-file-tile.tsx
stijnpotters1: No dragleave has no default action to cancel. It's just a notification that the cursor left the element. Our handleDragLeave only does bookkeeping, it decrements the depth counter and clears the highlight:

philipsens:
It might be interesting to explore the possibility to give this a more flat and calm design.

philipsens:
I changed it to a flat, calmer design:

image

PS: I also fixed the hidden sidebar that could still be opened. And the color of the sidebar. Check commits for more info.

philipsens (changes requested):
The border doesn't change back after dropping an element.

Also the add component modal looks very different from the one in the Studio.

stijnpotters1:
> The border doesn't change back after dropping an element.
> 
> Also the add component modal looks very different from the one in the Studio.

I have implemented these last tasks in the pr!

philipsens:
I think it is good. Do you think it would be hard to only show the forward of the hovered pipe to an exit and all when hovering over the exit?

stijnpotters1:
> I think it is good. Do you think it would be hard to only show the forward of the hovered pipe to an exit and all when hovering over the exit?

I will look into it! isn't this feature relevant for the element hiding pr?

philipsens:
> > I think it is good. Do you think it would be hard to only show the forward of the hovered pipe to an exit and all when hovering over the exit?
> 
> I will look into it! isn't this feature relevant for the element hiding pr?

Yeah wrong PR, had to much tabs open :)

But on second thought it was not needed. Maybe more of a configurable thing in the future or something.

philipsens (approved):

src/main/frontend/app/components/dialog.tsx
```diff
@@ -0,0 +1,31 @@
...
+type DialogProperties = {
+  isOpen: boolean
+  onClose: () => void
+  title: string
+  children: ReactNode
+  className?: string
+  overlay?: ReactNode
+}
+export default function Dialog({ isOpen, onClose, title, children, className, overlay }: Readonly<DialogProperties>) {
+  if (!isOpen) return null
```
Matthbo: Instead of isOpen property I rather see this component wrapped in an if statement

src/main/frontend/app/routes/configurations/add-non-canvas-component-menu.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { getAddableNonCanvasComponentNames } from '~/services/non-canvas-component-service'
+
+type AddNonCanvasComponentMenuProperties = {
+  isOpen: boolean
+  onClose: () => void
+  onSelect: (tagName: string) => void
+}
+
+export default function AddNonCanvasComponentMenu({
+  isOpen,
```
Matthbo: Also switch isOpen to if statement here

src/main/frontend/app/routes/configurations/non-canvas-component-context.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+    const directAttributes = element.attributes ?? {}
+    if (!ffDoc) return directAttributes
+
+    const inherited = getInheritedProperties(element, ffDoc.elements, ffDoc.enums)
+    return {
+      ...flattenInheritedParentElementProperties(inherited.attributesOptional),
+      ...flattenInheritedParentElementProperties(inherited.attributesRequired),
+      ...directAttributes,
+    }
```
Matthbo, line 65: I feel like doc-library has a function for this, but since im not fully sure I can take a look at it at some later point

src/main/frontend/app/routes/configurations/use-non-canvas-components.ts
```diff
@@ -0,0 +1,48 @@
...
+  replaceComponents: (configurationPath: string, components: NonCanvasComponent[]) => void
+}
+
+export function useNonCanvasComponents(projectName: string, configurationPaths: string[]): NonCanvasComponentsState {
+  const [componentsByPath, setComponentsByPath] = useState<Record<string, NonCanvasComponent[]>>({})
+  const [loadingByPath, setLoadingByPath] = useState<Record<string, boolean>>({})
+  const pathsKey = JSON.stringify(configurationPaths)
+
+  useEffect(() => {
+    const paths: string[] = JSON.parse(pathsKey)
```
Matthbo: Isn't this a little strange?

Matthbo (changes requested):
Also take a look at the linting warnings

src/main/frontend/app/routes/configurations/non-canvas-component-context.tsx
philipsens: Well did ya?

Matthbo: The Reference has a similar but different setup for this, but I think its fine to leave here if this works

Matthbo (approved):

PR #575, opened 2026-06-24
Refactored to original monaco editor instead of the /react version

### LINKED ISSUE #573: Monaco editor infinite loading
https://www.jsdelivr.com/package/npm/@monaco-editor/react
is loaded through a cdn and is added with <script> to the project at runtime
Now its blocked because of the new csp filter

stijnpotters1:
Now the monaco editor is bundled instead of loading the monaco editor from a CDN


Matthbo:
There still seems to be something that adds a monaco loader script to the dom in order to load the monace editor, instead of using the bundled assests
image

~~If this is the monaco bundled loader doing that it would be a little disappointing, but in case it isn't could you look for the source of this and preferably bundle the loader instead of doing any cdn work?~~

Matthbo (approved):

PR #587, opened 2026-06-30
Enhance flow handling by integrating forwards dynamically

### LINKED ISSUE #576: Gather the right forwards from the FrankDoc.json
I saw that a XsltSwitchPipe had a success forward. This is not possible since the pipe does not inherit from the FixedForwardPipe (the one that has success).

Image

This needs to be fixed.

stijnpotters1:
Now i have made logic to make different colors for each handle type.
The thing is that we could reuse the colour logic of #536 here as well. But this needs to be merged first
image
image

### DIFF
--- a/src/main/frontend/app/routes/studio/canvas/nodetypes/components/handle.tsx
+++ b/src/main/frontend/app/routes/studio/canvas/nodetypes/components/handle.tsx
@@ -24,10 +24,12 @@ const SEMANTIC_COLOURS: Record<string, string> = {
 const GREEN_BAND_START = 95
 const GREEN_BAND_SIZE = 70
 
+const HASH_MULTIPLIER = 31
+
 function colourFromName(type: string): string {
   let hash = 0
   for (const character of type) {
-    hash = (hash * 31 + (character.codePointAt(0) ?? 0)) % (360 - GREEN_BAND_SIZE)
+    hash = (hash * HASH_MULTIPLIER + (character.codePointAt(0) ?? 0)) % (360 - GREEN_BAND_SIZE)
   }
   const hue = hash < GREEN_BAND_START ? hash : hash + GREEN_BAND_SIZE
   return `hsl(${hue}, 65%, 52%)`
--- a/src/main/frontend/app/routes/studio/canvas/nodetypes/components/handle.tsx
+++ b/src/main/frontend/app/routes/studio/canvas/nodetypes/components/handle.tsx
@@ -21,35 +21,20 @@ const SEMANTIC_COLOURS: Record<string, string> = {
   error: '#ff7605ff',
 }
 
-const FORWARD_COLOURS: Record<string, string> = {
-  notfound: '#1B97D1',
-  empty: '#26A69A',
-  // eslint-disable-next-line unicorn/no-thenable
-  then: '#7E57C2',
-  else: '#EC407A',
-  lessthan: '#5C6BC0',
-  greaterthan: '#00ACC1',
-  equals: '#AB47BC',
-  stop: '#D81B60',
-  continue: '#42A5F5',
-  notinrole: '#8D6E63',
-  custom: '#9575CD',
-}
-
-const FALLBACK_PALETTE = Object.values(FORWARD_COLOURS)
+const GREEN_BAND_START = 95
+const GREEN_BAND_SIZE = 70
 
 function colourFromName(type: string): string {
   let hash = 0
-  for (let index = 0; index < type.length; index++) {
-    hash = (hash * 31 + (type.codePointAt(index) ?? 0)) % FALLBACK_PALETTE.length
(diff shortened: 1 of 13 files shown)


src/main/frontend/app/routes/studio/canvas/nodetypes/components/handle.tsx
```diff
@@ -13,25 +13,34 @@ interface HandleProperties {
...
-  default: '#1B97D1',
+}
+
+const GREEN_BAND_START = 95
+const GREEN_BAND_SIZE = 70
+
+function colourFromName(type: string): string {
+  let hash = 0
+  for (const character of type) {
+    hash = (hash * 31 + (character.codePointAt(0) ?? 0)) % (360 - GREEN_BAND_SIZE)
```
Matthbo: So why a base value of 31?
I remember seeing this done before and I questioned it back then too, making it a const and having a descriptive name would help a lot

src/main/frontend/app/utils/frankdoc-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function resolveForwardsWithInheritance(
+  element: ForwardSource | null | undefined,
+): Record<string, ElementProperty> {
+  const forwards = element?.forwards ?? {}
+
+  if (element?.labels?.EIP === 'Router') return forwards
+
+  return { success: {}, ...forwards }
```
Matthbo: Why add the success forward for all other elements?
If inheritance is handled correctly then every element that inherits from FixedForwardPipe should have the sucess forward which is defined on there

src/main/frontend/app/utils/frankdoc-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function getDefaultSourceHandles(resolvedForwards: Record<string, ElementProperty> | undefined): SourceHandle[] {
```
Matthbo:
```suggestion
export function getDefaultSourceHandles(resolvedForwards?: Record<string, ElementProperty>): SourceHandle[] {
```

Matthbo (changes requested):

philipsens:
The border is a bit weird. Can you fix that? 
image

src/main/frontend/app/routes/studio/canvas/nodetypes/components/handle.tsx
stijnpotters1: Because this is a prime number and is often used for multiple known algorithms like java.hashCode to generate good varying results. So in short its just to generate number to get a high enthropy

stijnpotters1: I made it a const

Matthbo:
I prefer if you didn't force-push when a PR is already in review, this makes it much more difficult to review

src/main/java/org/frankframework/flow/frankdoc/FrankDocService.java
```diff
@@ -13,22 +21,72 @@ public class FrankDocService {
...
+					findSuccessDeclaringAbstract(elements, siblings)
+							.ifPresent(successBase -> repointSiblingsToBase(elements, siblings, successBase))
+			);
+
+			return objectMapper.writeValueAsString(root);
+		} catch (Exception e) {
+			log.warn("Failed to repair FrankDoc parent chains, returning raw JSON", e);
+			return frankDocJson;
+		}
+	}
```
Matthbo, line 62: Does this imply an issue with the frankdoc element hierarchy?

src/main/frontend/vitest.config.ts
Matthbo, line 1: Should be fixed in the latest `@frankframework/doc-library-core` version
So this hopefully isn't needed

src/main/frontend/app/utils/frankdoc-utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  const inherited =
+    allElements && element.parent
+      ? getInheritedProperties(element as ElementClass, allElements, enums ?? {}).forwards
+      : {}
+
+  const merged = { ...inherited, ...ownForwards }
+
+  const isRouter = element.labels?.EIP === 'Router'
+  if (isRouter) {
+    const { success: _omit, ...withoutSuccess } = merged
```
Matthbo, line 34: What does _omit mean here?
Also I I think I commented on this before but success should be inherited from FixedForwardPipe.
Issue https://github.com/frankframework/frank-doc/issues/466 of the frank-doc needs to be fixed and this code updated to accurately have the success forward on the correct elements

philipsens:
I've looked trough the code, and apart from the color generation the code seems a bit redundant. I'll close this, so all usefull parts can be implemented in a new PR.

PR #591, opened 2026-06-30
Refactored zoom threshold logic and improve compact label rendering in nodes

### LINKED ISSUE #579: Extra readability in zoomed-out mode
The idea of the zoomed-out mode is so everything stays clear when wanting to see the big picture.

This is what it looks like before switching modes: 

Image

Not very readable, maybe you can make it out if you squint a bit. I think it could activate a bit earlier.

Also, when it switches, the forwards keep small:

Image

When zooming out it is even harder to read, while the big picture gets more clear:

Image

* I would suggest making the text scale a bit with the zoom level, if possible. So texts keeps readable even when completely zoomed out.
* I would remove the attribute values. It misses some information and can get ugly with long values. I don't think they improve anything. So just name would probably be best.

stijnpotters1:
image

### DIFF
--- a/src/main/frontend/app/routes/studio/canvas/nodetypes/zoomed-out-node.tsx
+++ b/src/main/frontend/app/routes/studio/canvas/nodetypes/zoomed-out-node.tsx
@@ -1,15 +1,14 @@
-import { Handle, Position } from '@xyflow/react'
-import { FlowConfig } from '~/routes/studio/canvas/flow.config'
+import { Handle, Position, useStore } from '@xyflow/react'
+import { FlowConfig, getCompactLabelScale } from '~/routes/studio/canvas/flow.config'
 
 const COMPACT_INITIALS_BOX_SIZE = 160
 const COMPACT_PADDING_TOP = 8
 const COMPACT_HANDLE_SIZE = 15
 const COMPACT_HANDLE_GAP = 4
+const COMPACT_LABEL_BASE_FONT_PX = 22
 
 export type ZoomedOutNodeProps = {
   subtype: string
-  name?: string
-  attributes?: Record<string, string>
   colorVariable: string
   selected?: boolean
   showTargetHandle?: boolean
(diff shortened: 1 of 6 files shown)


stijnpotters1:
@philipsens What do you think of this? Now the labels auto scale with zooming

Matthbo (approved):

philipsens:
I'll check it out and test it :)

src/main/frontend/app/routes/studio/canvas/nodetypes/zoomed-out-node.tsx
```diff
@@ -65,15 +63,9 @@ export default function ZoomedOutNode({
...
-        <span className="text-center text-3xl leading-snug font-semibold whitespace-nowrap">{subtype}</span>
-
-        {name && <span className="text-foreground-muted text-center text-3xl whitespace-nowrap">{name}</span>}
-        {attributes &&
-          Object.entries(attributes).map(([key, value]) => (
-            <span key={key} className="text-foreground-muted text-center text-2xl whitespace-nowrap">
-              {value || key}
-            </span>
-          ))}
+        <span className="text-center leading-snug font-semibold whitespace-nowrap" style={{ fontSize: labelFontSize }}>
```
philipsens:
```suggestion
        <span className="text-center leading-snug font-medium whitespace-nowrap" style={{ fontSize: labelFontSize }}>
```

philipsens (changes requested):
The stationary text feels nice and readable. But it could be less bold (medium instead of semibold). Also, I think the forwards should also scale with the zooming, so it is possible to see what is happening here:

Image

src/main/frontend/app/routes/studio/canvas/nodetypes/zoomed-out-node.tsx
stijnpotters1: I have applied it!

philipsens (approved):
Nice! Feels natural. Most people won't even know it is helping them haha

PR #595, opened 2026-07-01
Added hide/show functionality for selected nodes in the flow editor

### LINKED ISSUE #578: Make it possible to hide certain pipes for clarity
Today I was looking at some adapters and I saw that the forwards can become a lot when every pipe as an error flow.

Here is an example:

Image
(DetectRolChanges - Zaakbrug Translate)

It would be nice if there could be an `ff:hidden="true"` attribute, that hides the forwards and let the node appear a bit transparent. Hovering over it, would show all the forwards again.

Then the flow would look a lot clearer:

Image

stijnpotters1:
toggle on
image
on hover: 
image
toggle off
image

### DIFF
--- a/src/main/frontend/app/routes/studio/canvas/edgetypes/frank-edge.tsx
+++ b/src/main/frontend/app/routes/studio/canvas/edgetypes/frank-edge.tsx
@@ -14,6 +14,7 @@ export type FrankEdgeProperties = {
   selected?: boolean
   sourceHandleId?: string | null
   targetHandleId?: string | null
+  data?: { label?: string; faded?: boolean }
   type: string
 }
 
@@ -28,8 +29,10 @@ export default function FrankEdge({
   targetPosition,
   selected,
   sourceHandleId,
+  data,
 }: Readonly<FrankEdgeProperties>) {
   const deleteEdge = useFlowStore((state) => state.deleteEdge)
+  const faded = data?.faded ?? false
 
   const sourceHandleType = useFlowStore((state) => {
     const node = state.nodes.find((n) => n.id === source)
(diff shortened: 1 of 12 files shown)


philipsens:
Nice! How does it look in the zoomed-in mode? Do the forward connectors also change in opacity?

src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
```diff
@@ -95,6 +97,34 @@ export default function FrankNode(properties: NodeProps<FrankNodeType>) {
...
+  )
+
+  const dimmedHandleIndices = useMemo(() => {
+    const dimmed = new Set<number>()
+    if (hiddenForwardNodeIds.length === 0) return dimmed
+    const hiddenSet = new Set(hiddenForwardNodeIds)
+    const isRevealed = (targetId: string) =>
+      showAllForwards ||
+      hoveredNodeId === targetId ||
+      (hoveredNodeId !== null && edges.some((edge) => edge.source === hoveredNodeId && edge.target === targetId))
```
Matthbo: Maybe should be its own function

src/main/frontend/app/routes/studio/canvas/edgetypes/frank-edge.tsx
```diff
@@ -61,13 +64,20 @@ export default function FrankEdge({
 
   return (
     <>
-      <BaseEdge id={id} path={edgePath} style={{ strokeWidth: 3 }} />
+      <BaseEdge
+        id={id}
+        path={edgePath}
+        style={{ strokeWidth: 3, opacity: faded ? 0 : 1, transition: 'opacity 150ms cubic-bezier(0.4, 0, 0.2, 1)' }}
+        interactionWidth={faded ? 0 : undefined}
```
Matthbo, line 71: Does null also work instead of undefined?

stijnpotters1:
> Nice! How does it look in the zoomed-in mode? Do the forward connectors also change in opacity?

In zoomed in mode it looks like this:

image

src/main/frontend/app/routes/studio/canvas/edgetypes/frank-edge.tsx
stijnpotters1: No null would not work here. the default interactionWidth is 20 and when it becomes null instead of undefined its not falling back to the default width of 20.

src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
stijnpotters1: Fixed it!

src/main/frontend/app/routes/studio/canvas/edgetypes/frank-edge.tsx
Matthbo: Sad, stubborn library

Matthbo (approved):
Looks good to me, can be merged if Sergi agrees :D

philipsens:
I'm going to check it out!

philipsens:
It's working nice and intuitively. It might be nice to be able to hover on a targeting node to. 

Also: I expected that an exit also had the option to hide the forwards, but it doesn't.

Hiding the exception exit would be very useful for configurations like this one:

image

stijnpotters1:
> It's working nice and intuitively. It might be nice to be able to hover on a targeting node to.
> 
> Also: I expected that an exit also had the option to hide the forwards, but it doesn't.
> 
> Hiding the exception exit would be very useful for configurations like this one:
> 
> image

I have improved it!
Please let me know what you think of this, so i can merge it
Image

philipsens:
It's perfect :)

PR #607, opened 2026-07-08
Solved two small flow studio bugs

### LINKED ISSUE #604: frank node selection is unpredictable
Sometimes its selecting and sometimes not when there is already one selected

### LINKED ISSUE #606: Child adding does not open chosen child attributes in sidebar
Adding a child to a frank element does not open the attributes of this child for some elements

Right after adding a forward for example:

Image

stijnpotters1:
- Improved react flow node click that did not work always
- Fixed frank element child add sidebar opening

### DIFF
--- a/src/main/frontend/app/components/flow/add-subcomponent-modal.tsx
+++ b/src/main/frontend/app/components/flow/add-subcomponent-modal.tsx
@@ -60,7 +60,13 @@ export default function AddSubcomponentModal({ onClose, possibleChildren, onAddC
   useSubmitOnEnter(handleAddChild)
 
   return createPortal(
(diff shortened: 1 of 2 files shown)


src/main/frontend/app/components/flow/add-subcomponent-modal.tsx
Matthbo, line 1: Could you also cleanup the tailwind classes to something more sensible, such as z-[60] and min-w-[400px] to their actual tailwind values

philipsens: Done 👍🏻

philipsens (approved):

Matthbo (approved):

Issue #613, opened 2026-07-17
Rework implicit forwards and receiver to first pipe in pipeline


philipsens:
Currently everything gets a implicit forward in the xml-to-json-parser. That is the wrong place. Creating a new node in the canvas does not create a new implicit forward. Only when you close and open the adapter it gets shown. 

IfPipe also gets implicit forward. It is a bit strange because it isn't shown until a handle gets created, than it shows up. So it looks like you can have an implicit exception forward, which is not the case in this instance.

It should be dashed (like in design!!)

It should not be possible to change the implicit forward on the receiver. Maybe that should be another issue.

PR #614, opened 2026-07-17
Simplify adding success forwards to pipes

### LINKED ISSUE #576: Gather the right forwards from the FrankDoc.json

philipsens:
Now the success only gets added in the frank-node, instead of adding it on every node and removing it sometimes. 

The receivers now don't have a success (which is correct).

I discovered a new bug and made an issue for it: #613


src/main/frontend/app/routes/studio/canvas/nodetypes/frank-node.tsx
```diff
@@ -94,20 +97,42 @@ export default function FrankNode(properties: NodeProps<FrankNodeType>) {
...
 
-    return Object.values(recordElements).find((element) => element.name === properties.data.subtype) ?? null
-  }, [elements, properties.data.subtype])
+    return element
+  }, [elements, ffDoc, properties.data.subtype])
+
+  useEffect(() => {
+    if (!frankElement?.forwards) return
+
+    // console.log(frankElement.name, Object.keys(frankElement.forwards), sourceHandles.map((handle) => handle.type))
```
stijnpotters1:
```suggestion
```

stijnpotters1 (commented):
Should this be merged after https://github.com/frankframework/flow/issues/613 is implemented and combined with this or can these be merged seperately to master?

This because the source handles are commeted out currently

Matthbo:
I think the point of this PR was to at least partially cover #613 as well.
While working on this PR we came to the conclusion that removing complex code actually made things work more as they should rather than break things completely.
This is probably why its done like this and can be merged without worrying too much about the impact to other issues

Matthbo (approved):

